PR #106, opened 2024-03-25
Add one-dimensional Fermi-Hubbard model

### LINKED ISSUE #102: Add Fermi-Hubbard model
Create a new file `operators/fermi_hubbard.py`. In that file, define a function called `fermi_hubbard` with appropriate arguments. The function should return a [FermionOperator](https://github.com/qiskit-community/ffsim/blob/main/docs/tutorials/06-fermion-operator.ipynb).

bartandrews:
Fixes #102


bartandrews:
I have fixed the lint and format errors (as per the readme instructions), and the tox runs successfully on my computer now

.gitignore
```diff
@@ -105,6 +105,7 @@ celerybeat.pid
 # Environments
 .env
 .venv
+.ffsim
```
kevinsung, line 108: Please revert this change, since it seems to be special to your setup. You can add your own git ignore rules to `~/.config/git/ignore` (at least on Linux).

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,79 @@
...
+from ffsim.operators.fermion_action import cre_a, cre_b, des_a, des_b
+
+
+def fermi_hubbard(
+    norb: int,
+    t: float = 1,
+    U: float = 0,
+    mu: float = 0,
+    V: float = 0,
+    PBC: bool = False,
```
kevinsung, line 23: Instead of single-letter name or acronyms, give these variables descriptive names.

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,79 @@
...
+        mu: The chemical potential.
+        V: The nearest-neighbor interaction strength.
+        PBC: The periodic boundary conditions flag.
+
+    Returns:
+        The Fermi-Hubbard operator for a specified number of orbitals.
+    """
+    coeffs: dict[tuple[tuple[bool, bool, int], ...], complex] = {}
+
+    for p in range(norb - int(not PBC)):
```
kevinsung, line 52: I find `range(norb - 1 + PBC)` to be more readable.

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,79 @@
...
+        coeffs[(cre_a((p + 1) % norb), des_a(p))] = -t
+        coeffs[(cre_b((p + 1) % norb), des_b(p))] = -t
+        coeffs[(cre_a(p), des_a((p + 1) % norb))] = -t
+        coeffs[(cre_b(p), des_b((p + 1) % norb))] = -t
+        if V != 0:
```
kevinsung, line 57:
```suggestion
        if V:
```

kevinsung: Same comment below.

kevinsung (changes requested):
Thank you! Please add some tests.

bartandrews:
Thank you for the code review! I have implemented your suggestions and have added some tests, which benchmark the ground-state energies for an L=4 chain against known results (from pyblock3, TeNPy, etc.). The tox runs successfully on my computer

python/ffsim/operators/__init__.py
```diff
@@ -32,4 +33,5 @@
     "des_a",
     "des_b",
     "number_operator",
+    "fermi_hubbard",
```
kevinsung, line 36: Alphabetical order please.

python/ffsim/operators/fermi_hubbard.py
kevinsung: The new names still aren't descriptive enough. Suggestions:
- Instead of `t_hop`, `tunneling`
- Instead of `U_int`, `interaction`
- Instead of `mu_pot`, `chemical_potential`
- Instead of `V_pot`, `nearest_neighbor_interaction`
- Instead of `PBC`, `periodic`

You can also look at the corresponding function from [OpenFermion](https://github.com/quantumlib/OpenFermion/blob/e2fe1e678d973303fab231ce9b3cd83757413b94/src/openfermion/hamiltonians/hubbard.py#L20) for inspiration.

Also, let's not set default values for `t` and `U`.

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,79 @@
...
+
+def fermi_hubbard(
+    norb: int,
+    t_hop: float = 1,
+    U_int: float = 0,
+    mu_pot: float = 0,
+    V_int: float = 0,
+    PBC: bool = False,
+) -> FermionOperator:
+    r"""Fermi-Hubbard operator.
```
kevinsung, line 25:
```suggestion
    r"""Fermi-Hubbard model Hamiltonian.
```

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,79 @@
...
+
+    The Fermi-Hubbard operator for :math:`N` spatial orbitals is defined as
```
kevinsung, line 27:
```suggestion
    The Hamiltonian for the Fermi-Hubbard model with :math:`N` spatial orbitals is given by
```

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,79 @@
...
+    Args:
+        norb: The number of spatial orbitals.
+        t_hop: The hopping strength.
+        U_int: The onsite interaction strength.
+        mu_pot: The chemical potential.
+        V_int: The nearest-neighbor interaction strength.
+        PBC: The periodic boundary conditions flag.
+
+    Returns:
+        The Fermi-Hubbard operator for a specified number of orbitals.
```
kevinsung, line 48:
```suggestion
        The Fermi-Hubbard model Hamiltonian.
```

tests/operators/fermi_hubbard_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+def test_non_interacting_FH_operator():
+    """Test non-interacting Fermi-Hubbard operator."""
+
+    # open boundary conditions
+    op = fermi_hubbard(norb=4, t_hop=1)
+    ham = ffsim.linear_operator(op, norb=4, nelec=(2, 2))
+    eigs, _ = scipy.sparse.linalg.eigsh(ham, which="SA", k=1)
+    E = -4.472135955000
+    assert np.isclose(eigs[0], E)
```
kevinsung, line 32: Instead of
```python
assert np.isclose(a, b)
```
do
```
np.testing.assert_allclose(a, b)
```
This gives more informative error messages.

tests/operators/fermi_hubbard_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def test_non_interacting_FH_operator():
+    """Test non-interacting Fermi-Hubbard operator."""
+
+    # open boundary conditions
+    op = fermi_hubbard(norb=4, t_hop=1)
+    ham = ffsim.linear_operator(op, norb=4, nelec=(2, 2))
+    eigs, _ = scipy.sparse.linalg.eigsh(ham, which="SA", k=1)
+    E = -4.472135955000
```
kevinsung, line 31: Variable names should be all lowercase and descriptive (i.e. `energy` instead of `e`. In this case, I would avoid defining a variable altogether and just put the number directly in the assert statement.

kevinsung (changes requested):

bartandrews:
Thanks again for the suggested improvements -- I have implemented these changes now

bartandrews:
Thanks for catching this, I have implemented these changes now. Also, I want to confirm that in the docstring, when I write `` :math: `\mu` ``, is it okay to use to use a single slash here? Sometimes the backslash needs to be escaped/doubled for proper rendering

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,81 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Fermi-Hubbard model Hamiltonian."""
+
+from ffsim._lib import FermionOperator
+from ffsim.operators.fermion_action import cre_a, cre_b, des_a, des_b
+
+
+def fermi_hubbard(
```
kevinsung, line 17: It just occurred to me that this is only the one-dimensional Fermi-Hubbard model. I might need to rethink the API, because we definitely want to support the 2d model at some point. Let's rename this function to `fermi_hubbard_1d` for now.

kevinsung (changes requested):

bartandrews:
Thank you for adding a note on how to compile the docs locally. The mu was indeed rendered correctly, but viewing the docs allowed me to fix a different issue, which was useful. I have renamed to `fermi_hubbard_1d` in anticipation of a `fermi_hubbard_2d` in `fermi_hubbard.py` and corresponding functions `test_fermi_hubbard_2d...` in `fermi_hubbard_test.py`. I can implement that in a separate PR if needed

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,81 @@
...
+
+    .. math::
+
+        H = -t \sum_{p=1}^{N-1} \sum_{\sigma}
+        (a^\dagger_{p+1, \sigma} a_{p, \sigma} + \text{H.c.})
+        + \frac{U}{2} \sum_{p=1}^{N} \sum_{\sigma, \sigma'}
+        a^\dagger_{p, \sigma} a^\dagger_{p, \sigma'} a_{p, \sigma'} a_{p, \sigma}
+        - \mu \sum_{p=1}^N \sum_{\sigma} a^\dagger_{\sigma, p} a_{\sigma, p}
+        + V \sum_{p=1}^{N-1} \sum_{\sigma, \sigma'}
+        a^\dagger_{p, \sigma} a^\dagger_{p+1, \sigma'} a_{p+1, \sigma'} a_{p, \sigma}
```
kevinsung, line 38: Please rewrite the interaction terms by defining the number operator
```
n_{\sigma, p} = a_{\sigma, p}^\dagger a_{\sigma, p}
```
I think that will make it a bit more readable. Once you do that, make sure the order of the operators in the code matches the order in the documentation. Also, please consistently use (spin, index), i.e. `\sigma, p` instead of `p, \sigma`.

Also, I know the Hubbard model is very common, but it would still be good to have a reference for it.

kevinsung (changes requested):

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,81 @@
...
+
+def fermi_hubbard_1d(
+    norb: int,
+    tunneling: float,
+    interaction: float,
+    chemical_potential: float = 0,
+    nearest_neighbor_interaction: float = 0,
+    periodic: bool = False,
+) -> FermionOperator:
+    r"""1D Fermi-Hubbard model Hamiltonian.
```
kevinsung, line 25:
```suggestion
    r"""One-dimensional Fermi-Hubbard model Hamiltonian.
```

tests/operators/fermi_hubbard_test.py
```diff
@@ -0,0 +1,165 @@
...
+    ham_periodic = ffsim.linear_operator(op_periodic, norb=4, nelec=(1, 3))
+    eigs_periodic, _ = scipy.sparse.linalg.eigsh(ham_periodic, which="SA", k=1)
+    np.testing.assert_allclose(eigs_periodic[0], -0.828427124746)
+
+
+def test_fermi_hubbard_1d_hermiticity():
+    """Test 1D Fermi-Hubbard model Hamiltonian hermiticity."""
+
+    n_orbitals, n_alpha, n_beta = 4, 3, 1
+    dim = comb(n_orbitals, n_alpha) * comb(n_orbitals, n_beta)
```
kevinsung, line 138:
```suggestion
    norb = 4
    nelec = (3, 1)
    dim = ffsim.dim(norb, nelec)
```
Sorry for the extreme nitpicking but this is how it's done in the rest of the codebase.

tests/operators/fermi_hubbard_test.py
```diff
@@ -0,0 +1,165 @@
...
+    # open boundary conditions
+    op = fermi_hubbard_1d(
+        norb=n_orbitals,
+        tunneling=0.1,
+        interaction=0.2,
+        chemical_potential=0.3,
+        nearest_neighbor_interaction=0.4,
+    )
+    ham = ffsim.linear_operator(op, norb=n_orbitals, nelec=(n_alpha, n_beta))
+    np.testing.assert_allclose(ham.dot(np.eye(dim)), ham.H.dot(np.eye(dim)))
```
kevinsung, line 149: Instead of `a.dot(b)`, do `a @ b`, it's more readable and consistent with the rest of the codebase.

kevinsung (changes requested):
The eigenvalue tests are great but I'm trying to think of a test that's more "direct". I think this might work:
1. Create a Fermi-Hubbard model
2. Print its string representation using `repr` and manually check that it makes sense
3. In the test, assert that the string representation is equal to the one obtained in step 2.

bartandrews:
Thank you for catching these inconsistencies, I have implemented the changes now :rocket: The next pull request should be smoother once I have learnt your code conventions :+1: (I have also changed the test location to be in the python subdirectory.)

tests/python/operators/fermi_hubbard_test.py
```diff
@@ -0,0 +1,251 @@
...
+
+    # open boundary conditions
+    op = fermi_hubbard_1d(norb=4, tunneling=1, interaction=0)
+    ham = ffsim.linear_operator(op, norb=4, nelec=(2, 2))
+    coeffs: dict[tuple[tuple[bool, bool, int], ...], complex] = {}
+    for p in range(3):
+        coeffs[(cre_a(p), des_a((p + 1) % 4))] = -1
+        coeffs[(cre_b(p), des_b((p + 1) % 4))] = -1
+        coeffs[(cre_a((p + 1) % 4), des_a(p))] = -1
+        coeffs[(cre_b((p + 1) % 4), des_b(p))] = -1
```
kevinsung, line 35: I had something a bit different in my mind for the test, actually. Please add a single test, similar to [this one](https://github.com/qiskit-community/ffsim/blob/561052c6486e7daf7f828a8d4096eff99f26cc33/tests/python/operators/common_operators_test.py#L21) for `number_operator`. The test should compare the dictionary of the operator with one constructed explicitly without any for loop logic. Then, you can revert these other tests back to how they were originally, where the only thing they did was check the eigenvalue. That gives a better logical separation between what the tests are doing. For good measure, you can rename the tests by appending `_eigenvalue` to the test name and mention it in the docstring, e.g.
```python
def test_non_interacting_fermi_hubbard_1d_eigenvalue():
    """Test non-interacting 1D Fermi-Hubbard model eigenvalues."""
```

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,87 @@
+# (C) Copyright IBM 2023.
```
kevinsung, line 1:
```suggestion
# (C) Copyright IBM 2024.
```

tests/python/operators/fermi_hubbard_test.py
```diff
@@ -0,0 +1,251 @@
+# (C) Copyright IBM 2023.
```
kevinsung, line 1:
```suggestion
# (C) Copyright IBM 2024.
```

kevinsung (changes requested):

bartandrews:
As requested, in the tests I have now "compared the dictionary of the operator with one constructed explicitly without any for loop logic", for open and periodic boundary conditions. In the process, I also noticed that some existing tests in ffsim use `assert A == B` rather than `np.testing.assert_equal(A, B)`, which seems to go against your convention. Should these also be fixed?

kevinsung:
>assert A == B rather than np.testing.assert_equal(A, B)

The reason to use `np.testing.assert_allclose` is that the error message displays the values of the arrays being compared. When A and B aren't arrays then the error messages from pytest are good enough. In any case, unrelated improvements should be done in a separate PR.

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,87 @@
...
+        interaction: The onsite interaction strength :math:`U`.
+        chemical_potential: The chemical potential :math:`\mu`.
+        nearest_neighbor_interaction: The nearest-neighbor interaction strength
+            :math:`V`.
+        periodic: Whether to use periodic boundary conditions.
+
+    Returns:
+        The one-dimensional Fermi-Hubbard model Hamiltonian.
+
+    .. _The Hubbard Model: https://doi.org/10.1146/annurev-conmatphys-031620-102024
```
kevinsung, line 56: Thanks for adding this reference. Do you happen to know of one that also includes the nearest-neighbor interaction? I don't think this one does, but perhaps I missed it. I actually haven't seen this term as part of the Hubbard model before.

bartandrews: For the Fermi-Hubbard model, I was following a similar functionality to TeNPy: https://tenpy.readthedocs.io/en/latest/reference/tenpy.models.hubbard.FermiHubbardModel.html#tenpy.models.hubbard.FermiHubbardModel However, the additional V term is fairly innocuous. For example, Eq.1 in the reference doesn't have a chemical potential term either, but this can simply be added, in case it's needed.

bartandrews: Sometimes this is referred to as the "extended Fermi-Hubbard model", e.g. https://arxiv.org/abs/2209.11777

bartandrews: However, it is difficult to find a single reference for this, because you can add on whatever you like to the Fermi-Hubbard model. The V nearest-neighbor term just seemed like it might be useful.

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -0,0 +1,87 @@
...
+
+from ffsim._lib import FermionOperator
+from ffsim.operators.fermion_action import cre_a, cre_b, des_a, des_b
+
+
+def fermi_hubbard_1d(
+    norb: int,
+    tunneling: float,
+    interaction: float,
+    chemical_potential: float = 0,
```
kevinsung, line 21:
```suggestion
    interaction: float,
    *,
    chemical_potential: float = 0,
```
Let's force keyword arguments here so we can add more later without breaking the API.

python/ffsim/operators/fermi_hubbard.py
kevinsung: I see, thanks. I think it's fine then.

python/ffsim/operators/fermi_hubbard.py
bartandrews: That's a good idea! I have added this now

kevinsung (approved):
Thanks Bart!

PR #142, opened 2024-04-24
Fix edge case in fermi_hubbard_1d

bartandrews:
This PR fixes an edge case in fermi_hubbard_1d, which appears when the number of spatial orbitals N=2, and periodic boundary conditions are set (i.e. a ring with two sites). In this case, the hopping due to the periodic boundary conditions has the same dictionary key as the internal hopping. Instead of this dictionary key being overwritten, it should be added.

### DIFF
--- a/python/ffsim/operators/fermi_hubbard.py
+++ b/python/ffsim/operators/fermi_hubbard.py
@@ -58,11 +58,18 @@ def fermi_hubbard_1d(
     """
     coeffs: dict[tuple[tuple[bool, bool, int], ...], complex] = {}
 
+    # initialize tunneling keys
     for p in range(norb - 1 + periodic):
-        coeffs[(cre_a(p), des_a((p + 1) % norb))] = -tunneling
-        coeffs[(cre_b(p), des_b((p + 1) % norb))] = -tunneling
-        coeffs[(cre_a((p + 1) % norb), des_a(p))] = -tunneling
-        coeffs[(cre_b((p + 1) % norb), des_b(p))] = -tunneling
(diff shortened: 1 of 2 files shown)


python/ffsim/operators/fermi_hubbard.py
```diff
@@ -58,11 +58,19 @@ def fermi_hubbard_1d(
     """
     coeffs: dict[tuple[tuple[bool, bool, int], ...], complex] = {}
 
+    # initialize tunneling keys
```
kevinsung, line 61: Let's make `coeffs` a [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict). Then, we won't need this block.

bartandrews: Thanks, well spotted! I have changed this for the fermi_hubbard_2d PR as well.

python/ffsim/operators/fermi_hubbard.py
```diff
@@ -56,13 +58,13 @@ def fermi_hubbard_1d(
 
     .. _The Hubbard Model: https://doi.org/10.1146/annurev-conmatphys-031620-102024
     """
-    coeffs: dict[tuple[tuple[bool, bool, int], ...], complex] = {}
+    coeffs: dict[tuple[tuple[bool, bool, int], ...], complex] = defaultdict(lambda: 0)
```
kevinsung, line 61: It should be
```python
defaultdict(float)
```
If mypy complains, then do
```python
defaultdict(complex)
```

kevinsung (changes requested):

kevinsung (approved):

PR #193, opened 2024-05-25
Add diagonal Coulomb Hamiltonian class

### LINKED ISSUE #117: Add DiagonalCoulombHamiltonian
This is a Hamiltonian with a diagonal Coulomb term. See Eq. (1) of https://arxiv.org/pdf/1711.04789.pdf. We can absorb the "number" term into the one body-part.

We can use the corresponding class in [OpenFermion](https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/ops/representations/diagonal_coulomb_hamiltonian.py) as a reference.

The implementation should be similar to [MolecularHamiltonian](https://github.com/qiskit-community/ffsim/blob/main/python/ffsim/hamiltonians/molecular_hamiltonian.py) or [DoubleFactorizedHamiltonian](https://github.com/qiskit-community/ffsim/blob/main/python/ffsim/hamiltonians/double_factorized_hamiltonian.py). Something like this:

```python
@dataclasses.dataclass(frozen=True)
class DiagonalCoulombHamiltonian:
    one_body_tensor: np.ndarray
    diag_coulomb_mat: np.ndarray
```

We can add a method
```python
@staticmethod
def from_fermion_operator(op: FermionOperator) -> DiagonalCoulombHamiltonian
```
to initialize from a compatible FermionOperator, like the Hubbard model. Passing a FermionOperator with incompatible terms should raise an error.

bartandrews:
Fixes #117


bartandrews:
This is a semi-draft pull request, i.e. I have a complete implementation, but I'm not sure if it's what was intended. Any guidance would be greatly appreciated! :smiley:

kevinsung (changes requested):

tests/python/hamiltonians/diagonal_coulomb_hamiltonian_test.py
```diff
@@ -0,0 +1,158 @@
...
+@pytest.mark.parametrize(
+    "norb, nelec",
+    [
+        (4, (2, 2)),
+        (4, (1, 2)),
+        (4, (0, 2)),
+        (4, (0, 0)),
+    ],
+)
+def test_from_fermion_operator(norb: int, nelec: tuple[int, int]):
```
kevinsung, line 108: Please add another test for `from_fermion_operator` where you initialize the DiagonalCoulombHamiltonian from the output of `ffsim.fermi_hubbard_1d` and `ffsim.fermi_hubbard_2d` and check that the resulting tensors are what you expect them to be.

bartandrews: I have added these tests and have separated them into tests for Fermi-Hubbard models with/without a U term. Since, according to my definition, the Hubbard U term is not compatible with the DiagonalCoulombHamiltonian.

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,195 @@
...
+                    (cre_a(p), cre_a(q), des_a(q), des_a(p)): coeff,
+                    (cre_a(p), cre_b(q), des_b(q), des_a(p)): coeff,
+                    (cre_b(p), cre_a(q), des_a(q), des_b(p)): coeff,
+                    (cre_b(p), cre_b(q), des_b(q), des_b(p)): coeff,
+                }
+            )
+        return op
+
+    @staticmethod
+    def from_fermion_operator(op: FermionOperator) -> DiagonalCoulombHamiltonian:
```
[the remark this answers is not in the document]
bartandrews: I have not normal ordered the operator anymore. Apart from checks for incompatible onsite interaction terms and incompatible leftover terms, we now simply iterate through and populate the tensors. One important detail is that at the end, we need to ensure that the `diag_coulomb_mat` is symmetric. This standardizes terms like n_i n_j = (n_i n_j + n_j n_i)/2

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,195 @@
...
+            (cre_b(p), des_a(q)),
+            (cre_a(p), des_b(q)),
+        ]
+        for key in excluded_list:
+            if key in dict_op:
+                del dict_op[key]
+
+        # check for incompatible terms
+        if dict_op:
+            print(f"Incompatible terms = {dict_op}")
```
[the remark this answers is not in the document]
bartandrews: Actually I left this on purpose because I thought that it might be helpful information for the user. However, I have removed it now

tests/python/hamiltonians/diagonal_coulomb_hamiltonian_test.py
kevinsung: Oh, I see. Well, we do want this class to be able to represent the Hubbard model. This means that we'll need to change the definition. Instead of a single diagonal Coulomb matrix, there should be three, for the alpha-alpha, alpha-beta, and beta-beta interactions. For the Hubbard model, the alpha-beta matrix will be diagonal with entries U, and the alpha-alpha and alpha-beta matrices will have nonzero terms for the nearest-neighbor interactions. And since we're making this generalization, we should also allow two separate one-body tensors, one for each spin (for the Hubbard model they will be identical). So we'll change
```python
    one_body_tensor: np.ndarray  # shape: (norb, norb)
    diag_coulomb_mat: np.ndarray  # shape: (norb, norb)
```
to
```python
    one_body_tensors: np.ndarray  # shape: (2, norb, norb)
    diag_coulomb_mats: np.ndarray  # shape: (3, norb, norb)
```

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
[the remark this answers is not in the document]
kevinsung: It's still not what I have in mind. Let me try to clarify a few points.
- There's no need to convert the operator to a dictionary. You can already iterate through it like a dictionary.
- There's no need to delete items.
- There shouldn't be any loops over `range(norb)` or its product with itself. There should be just one loop like this:
  ```python
  for term, coeff in op.items():
      # check if term is valid, and if so, populate the appropriate tensor
      # if the term is not valid, raise an error
  ```

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,213 @@
...
+                    diag_coulomb_mats[2][p, q] += 2 * np.real(coeff)
+                else:
+                    raise ValueError(
+                        "FermionOperator cannot be converted to DiagonalCoulombHamiltonian")
+            else:
+                raise ValueError("FermionOperator cannot be converted to DiagonalCoulombHamiltonian")
+
+        # ensure diag_coulomb_mats symmetry
+        for i in range(3):
+            diag_coulomb_mats[i] = (diag_coulomb_mats[i] + diag_coulomb_mats[i].T) / 2
```
kevinsung, line 197: Extremely minor nit but ffsim convention is to multiply by 0.5 instead of divide by 2. Multiplication is a simpler numerical operation than division and is preferred when the number can be represented exactly in binary.

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # print("op here = ", op)
+
+        # extract norb
+        orb_list = []
+        for key, _ in op.items():
+            for operator in key:
+                orb = operator[2]
+                orb_list.append(orb)
+        norb = max(orb_list) + 1
```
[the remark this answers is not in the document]
bartandrews: Thank you for these suggested improvements. I will implement these, and the other corrections, asap

tests/python/hamiltonians/diagonal_coulomb_hamiltonian_test.py
kevinsung: On further thought, there's probably no need to be so general. Let's just have two diagonal Coulomb matrices, one for same-spin and one for different spin. And also just a single one-body tensor. So like this:
```
    one_body_tensors: np.ndarray  # shape: (norb, norb)
    diag_coulomb_mats: np.ndarray  # shape: (2, norb, norb)
```

bartandrews: I have now redefined the DiagonalCoulombHamiltonian based on this 3rd definition. I also think that it's the best of the 3, especially if one of the aims is to get a compact representation of the Hubbard model.

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,171 @@
...
+        one_body_tensor = np.zeros((norb, norb), dtype=complex)
+        diag_coulomb_mats = np.zeros((2, norb, norb), dtype=float)
+
+        # populate the tensors
+        for term, coeff in op.items():
+            if not term:  # constant term
+                constant = coeff.real
+            elif len(term) == 2:  # one-body term candidate
+                (_, _, p), (_, _, q) = term
+                terms = [(cre_a(p), des_a(q)), (cre_b(p), des_b(q))]
```
kevinsung, line 131: Rename this variable to `valid_terms`.

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,171 @@
...
+                if term in terms:
+                    one_body_tensor[p, q] += 0.5 * coeff
+                else:
+                    raise ValueError(
+                        "FermionOperator cannot be converted to "
+                        "DiagonalCoulombHamiltonian"
```
kevinsung, line 137: Make this error message more descriptive. Explain that it encountered a quadratic term that was not of the form `a^\dagger_p a_q`, and include the term in the error message (use an f-string).

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,171 @@
...
+                    )
+            elif len(term) == 4:  # two-body term candidate
+                (_, _, p), (_, _, _), (_, _, q), (_, _, _) = term
```
kevinsung, line 140:
```suggestion
                (_, _, p), _, (_, _, q), _ = term
```

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,171 @@
...
+                    (cre_b(p), des_b(p), cre_a(q), des_a(q)),
+                ]
+                if term in terms_same_spin:
+                    diag_coulomb_mats[0][p, q] += coeff.real
+                elif term in terms_diff_spin:
+                    diag_coulomb_mats[1][p, q] += coeff.real
+                else:
+                    raise ValueError(
+                        "FermionOperator cannot be converted to "
+                        "DiagonalCoulombHamiltonian"
```
kevinsung, line 156: Same comment as above about making the error message more descriptive.

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,171 @@
...
+                    )
+            else:
+                raise ValueError(
+                    "FermionOperator cannot be converted to DiagonalCoulombHamiltonian"
```
kevinsung, line 160: As above, make this error message more descriptive.

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,171 @@
...
+                )
+
+        # ensure diag_coulomb_mats symmetry
+        for i in range(2):
+            diag_coulomb_mats[i] = 0.5 * (diag_coulomb_mats[i] + diag_coulomb_mats[i].T)
```
kevinsung, line 165: Use the `transpose` function from Numpy and do it in place to avoid allocating more memory than needed.
```suggestion
        diag_coulomb_mats += diag_coulomb_mats.transpose(0, 2, 1)
        diag_coulomb_mats *= 0.5
```

kevinsung (changes requested):
Looking much better, thank you Bart!

python/ffsim/hamiltonians/diagonal_coulomb_hamiltonian.py
bartandrews: Nice!

kevinsung (approved):
LGTM, thank you!

PR #199, opened 2024-05-26
Fix docstring for fermi_hubbard_1d

### LINKED ISSUE #186: Docstring for fermi_hubbard_1d doesn't accurately describe periodic boundary conditions
https://qiskit-community.github.io/ffsim/api/ffsim.html#ffsim.fermi_hubbard_1d

For example, the equation is only applicable for the case of non-periodic boundary conditions.

bartandrews:
Fixes #186


kevinsung (changes requested):
This still isn't quite right, because we'd need to specify that p+1 needs to be taken modulo n in the case of periodic boundary conditions. I think it might be better to rewrite it to be a bit more general, something like saying the sum runs over the edges in a line, and in the case of periodic conditions the line is connected back to itself as a ring, and that a ring with 2 nodes is considered to have 2 edges, etc.

bartandrews:
Ah, I see what you mean now. No problem, I can do that. In that case, should I also fix the `fermi_hubbard_2d` docstring in the same pull request?

kevinsung:
> Ah, I see what you mean now. No problem, I can do that. In that case, should I also fix the `fermi_hubbard_2d` docstring in the same pull request?

sure!

kevinsung:
Thanks Bart, I went ahead and made some final changes myself.

PR #247, opened 2024-06-21
Sample from Slater determinant

jrm874:
Adds `sample_slater()` function in the `ffsim.states` module. 

- `sample_slater()` allows to sample electronic configurations given the 1-RDM that defined the Slater determinant. 
- `sample_slater()` supports sampling for spineless and spinfull systems.
- The samples are generated using the Metropolis-Hastings algorithm


python/ffsim/states/slater.py
```diff
@@ -0,0 +1,326 @@
...
+    nelec = round(np.sum(np.diag(rdm)))
+
+    if nelec < n_particles_to_move:
+        raise ValueError(
+            "Number of electrons smaller than ``n_particles_to_move``: "
+            + "number of electrons ("
+            + str(nelec)
+            + ") and  ``n_particles_to_move`` ("
+            + str(n_particles_to_move)
+            + ")"
```
kevinsung, line 301: The `+`s are unnecessary. Also, please use f-strings. Something like this:
```suggestion
            "Number of electrons smaller than ``n_particles_to_move``: "
            f"number of electrons ({nelec}) "
            f"and ``n_particles_to_move`` ({n_particles_to_move})."
```

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+
+    _, u = la.eigh(rdm)
```
kevinsung, line 304: Rename `u` to `vecs`, it's more descriptive.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,326 @@
+import numpy as np
+import scipy.linalg as la
```
kevinsung, line 2: Nit: In ffsim we just do
```python
import scipy.linalg
```
without renaming it.

kevinsung (changes requested):
This needs some tests. You can use the newly introduced functions [`slater_determinant_rdms`](https://github.com/qiskit-community/ffsim/blob/6526eca81526c36d87e7bc96865fc5081337f595/python/ffsim/states/states.py#L281) and [`sample_state_vector`](https://github.com/qiskit-community/ffsim/blob/2a1b5e989db3223668173e0993e6ebbb19aceea0/python/ffsim/states/states.py#L564) in your tests (you'll need to merge main and fix conflicts).

I made a first pass with some minor comments on coding practices.

python/.DS_Store
kevinsung, line 1: This file doesn't belong.

tests/python/states/states_test.py
```diff
@@ -331,6 +332,90 @@ def test_indices_and_strings_roundtrip_spinful(norb: int, nelec: tuple[int, int]
...
+    n_a, n_b = nelec
+
+    rotation_a = ffsim.random.random_unitary(norb, seed=rng)
+    rotation_b = ffsim.random.random_unitary(norb, seed=rng)
+
+    vecs_a = rotation_a[:, :n_a]
+    vecs_b = rotation_b[:, :n_b]
+
+    rdm_a = vecs_a @ np.conjugate(vecs_a.T)
+    rdm_b = vecs_b @ np.conjugate(vecs_b.T)
```
kevinsung, line 372: Instead of computing these manually, use [`ffsim.slater_determinant_rdms`](https://qiskit-community.github.io/ffsim/api/ffsim.html#ffsim.slater_determinant_rdms).

tests/python/states/states_test.py
```diff
@@ -331,6 +332,90 @@ def test_indices_and_strings_roundtrip_spinful(norb: int, nelec: tuple[int, int]
...
+
+    test_distribution = (
+        np.absolute(
+            ffsim.slater_determinant(
+                norb,
+                (list(np.arange(n_a)), list(np.arange(n_b))),
```
kevinsung, line 378:
```suggestion
                (range(n_a), range(n_b)),
```

tests/python/states/states_test.py
```diff
@@ -331,6 +332,90 @@ def test_indices_and_strings_roundtrip_spinful(norb: int, nelec: tuple[int, int]
...
+    rotation_b = ffsim.random.random_unitary(norb, seed=rng)
+
+    vecs_a = rotation_a[:, :n_a]
+    vecs_b = rotation_b[:, :n_b]
+
+    rdm_a = vecs_a @ np.conjugate(vecs_a.T)
+    rdm_b = vecs_b @ np.conjugate(vecs_b.T)
+
+    test_distribution = (
+        np.absolute(
```
kevinsung, line 375:
```suggestion
        np.abs(
```
Maybe this will make the formatting less aggressive?

tests/python/states/states_test.py
```diff
@@ -331,6 +332,90 @@ def test_indices_and_strings_roundtrip_spinful(norb: int, nelec: tuple[int, int]
...
+    """Test Slater determinant sampler."""
+    rng = np.random.default_rng(1234)
+
+    n_samples = 3000
+
+    rotation = ffsim.random.random_unitary(norb, seed=rng)
+
+    vecs = rotation[:, :nelec]
+
+    rdm = vecs @ np.conjugate(vecs.T)
```
kevinsung, line 403: Same comment as above, compute these using `ffsim.slater_determinant_rdms`.

tests/python/states/states_test.py
```diff
@@ -331,6 +332,90 @@ def test_indices_and_strings_roundtrip_spinful(norb: int, nelec: tuple[int, int]
...
+
+    test_distribution = (
+        np.absolute(
+            ffsim.slater_determinant(norb, (list(np.arange(nelec)), []), rotation)
```
kevinsung, line 407: If you do the same changes I suggested above (`absolute` -> `abs` and just use `range`) maybe the formatting will be less aggressive.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+import scipy.linalg
+
+
+def sample_slater(
+    rdm: np.ndarray | tuple[np.ndarray, np.ndarray],
+    chain_length: int,
+    n_chains: int,
+    n_particles_to_move: int = 1,
+    seed=None,
+) -> np.ndarray:
```
kevinsung, line 13: For now, let's make the input and output formats match [`ffsim.sample_statevector`](https://qiskit-community.github.io/ffsim/api/ffsim.html#ffsim.sample_state_vector), which has this signature:
```python
def sample_state_vector(
    rdm: np.ndarray | tuple[np.ndarray, np.ndarray],
    *,
    orbs: list[int],
    shots: int,
    norb: int | None = None,
    nelec: int | tuple[int, int] | None = None,
    seed: np.random.Generator | int | None = None,
) -> list[str]:
```
So here, we would have
```python
def sample_slater(
    vec: np.ndarray | StateVector,
    *,
    orbs: list[int],
    shots: int,
    norb: int | None = None,
    nelec: int | tuple[int, int] | None = None,
    seed: np.random.Generator | int | None = None,
) -> list[str]:
```
For the other parameters you have, like `chain_length`, you can add those as extra parameters, but with some reasonable default values.

Note that the return type is a list of strings of length `shots`. This isn't the most efficient, but we can revisit this in the future if needed. Or, if you think it's important to allow the return type to be a bit array, perhaps we can handle that as an additional option, which we should probably define an enum for. Something like
```python
return_type: BitstringType = BitstringType.STRING   # enum representing either string, bit array, or integer
```

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+
+    Returns:
+        A 2D Numpy array with samples of electronic configurations.
+        Each row is a sample.
+    """
+
+    if isinstance(rdm, np.ndarray) and rdm.ndim == 2:
+        # spinless case
+        n = round(np.real(np.sum(np.diag(rdm))))
+        norb = rdm.shape[0]
```
kevinsung, line 39:
```suggestion
        norb, _ = rdm.shape
```
This has the advantage that it will throw an error on an improperly shaped input.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+
+        if n == 0 or n == norb:
```
kevinsung, line 41: Don't combine these cases. Use `np.zeros` in one case and `np.ones` in the other.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+            sampled_configuration = np.ones(
+                (n_chains * (chain_length + 1), rdm.shape[0]), dtype=int
```
kevinsung, line 43:
```suggestion
                (n_chains * (chain_length + 1), norb), dtype=int
```

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+        previous step: A Numpy array with the starting positions of the electrons.
+        norb: Number of orbitals.
+        n_particles_to_move: At each step of the chain, the maximum number of
+            electrons that are moved into a different orbital.
+
+    Returns:
+        A Numpy array with the proposed new set of positions of the electrons.
+    """
+
+    rng = np.random.default_rng(rng)
```
kevinsung, line 101: The `rng` should be initialized at the beginning of `sampler_slater`, and then passed to helper functions.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+    if isinstance(rdm, np.ndarray) and rdm.ndim == 2:
+        # spinless case
+        n = round(np.real(np.sum(np.diag(rdm))))
+        norb = rdm.shape[0]
+
+        if n == 0 or n == norb:
+            sampled_configuration = np.ones(
+                (n_chains * (chain_length + 1), rdm.shape[0]), dtype=int
+            ) * int(n / norb)
+
```
kevinsung, line 45: Nit: Your use of whitespace is a bit excessive by ffsim standards 😄

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+    for i in range(n_chains):
+        sample_particles_to_move = rng.choice(np.arange(n_particles_to_move) + 1, 1)
+
+        open_orbitals = np.setdiff1d(positions, previous_step[i], assume_unique=True)
+
+        electron_id = np.sort(
+            rng.choice(np.arange(nelec), sample_particles_to_move, replace=False)
+        )
+
```
kevinsung, line 118: Excessive whitespace.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+    Returns: Tuple. The first element is a Numpy array with the signs of the
+        determinants. The second element is a Numpy array with the logarithm
+        of the absolute value of the determinant.
+    """
+
+    det_matrices = phi[positions, :]
+
+    signs, logdet_amps = np.linalg.slogdet(det_matrices)
+
+    return signs, logdet_amps
```
kevinsung, line 152: This helper function seems unnecessary. Remove it and simply call `np.linalg.slogdet` directly instead.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+            configurations for each chain. The second element of the tuple is a
+            Numpy array with the logarithm of the absolute value of the
+            determinant of the previous configurations.
+
+    Returns:
+        A Numpy array of ``bool`` type that specifies if the move for each chain
+        is accepted (``True``) of rejected (``False``).
+    """
+
+    rng = np.random.default_rng(rng)
```
kevinsung, line 180: Same comment as above about the rng. It should be initialized only once in the main function.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+    """
+    Proposes initial positions for the electrons to initialize the Markov chains.
+    Electron positions are sampled from the diagonal part of the one-body
+    RDM.
+
+    Args:
+        n_chains: Number of chains.
+        nelec: Number of electrons.
+        norb: Number of orbitals.
+        rdm: A Numoy array with the one-body RDM.
```
kevinsung, line 207: Typo. `Numpy` not `Numoy`

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+
+    Returns:
+        A 2Dimensional Numpy array with samples of electronic configurations.
+        Each row is a sample.
+    """
+
+    rng = np.random.default_rng(seed)
+
+    norb = rdm.shape[0]
+
```
kevinsung, line 321: excessive whitespace

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+            f"number of electrons ({nelec}) "
+            f"and ``n_particles_to_move`` ({n_particles_to_move})."
+        )
+
+    _, vecs = scipy.linalg.eigh(rdm)
+
+    phi = vecs[:, -nelec:]
+
+    positions = np.zeros((n_chains, chain_length + 1, nelec), dtype=int)
+
```
kevinsung, line 336: excessive whitespace

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+    positions[:, 0, :] = _initialize_chains(n_chains, nelec, norb, rdm, rng)
```
kevinsung, line 337: Ending colon is always unnecssary, here and elsewhere.

python/ffsim/states/slater.py
```diff
@@ -0,0 +1,353 @@
...
+    rdm: np.ndarray | tuple[np.ndarray, np.ndarray],
+    chain_length: int,
+    n_chains: int,
+    n_particles_to_move: int = 1,
+    seed=None,
+) -> np.ndarray:
+    """
+    Collect samples of electronic configurations from a Slater determinant defined
+    by its one-body reduced density matrix (RDM). The sampler uses the
+    Metropolis-Hastings method.
```
kevinsung, line 17: The format of a docstring should be
```python
def func():
"""Single-line summary.

Main body with extended description. Can be multiple lines.
"""
```

In the main body of the docstring, please provide enough context or background information that the `chain_length` and other arguments can be understood. I see you wrote
>chain_length: Length of each sampling chain.

but you haven't explained what a sampling chain _is_.

kevinsung (changes requested):

PR #334, opened 2024-10-25
Add method to convert MolecularHamiltonian to MPO

bartandrews:
I have added a method to the `MolecularHamiltonian` class that can convert it to an MPO. Following the convention used for `qiskit`, I added a `tenpy` directory, with a `util.py` file. It also has a `hamiltonians` directory, anticipating other Hamiltonians to be implemented in the future.

### DIFF
--- a/python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
+++ b/python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
@@ -1,0 +1,111 @@
+import numpy as np
+from tenpy.models.lattice import Lattice
+from tenpy.models.model import CouplingMPOModel
+from tenpy.networks.site import SpinHalfFermionSite
+
+# ignore lowercase argument and variable checks to maintain TeNPy naming conventions
+# ruff: noqa: N803, N806
+
+
+class MolecularChain(Lattice):
+    def __init__(self, L, norb, site_a, **kwargs):
+        basis = np.array(([norb, 0.0], [0, 1]))
+        pos = np.array([[i, 0] for i in range(norb)])
+
+        kwargs.setdefault("order", "default")
+        kwargs.setdefault("bc", "open")
+        kwargs.setdefault("bc_MPS", "finite")
+        kwargs.setdefault("basis", basis)
+        kwargs.setdefault("positions", pos)
+
+        super().__init__([L, 1], [site_a] * norb, **kwargs)
+
+
+class MolecularHamiltonianMPO(CouplingMPOModel):
+    def __init__(self, params):
+        CouplingMPOModel.__init__(self, params)
+
+    def init_sites(self, params):
+        cons_N = params.get("cons_N", "N")
+        cons_Sz = params.get("cons_Sz", "Sz")
+        site = SpinHalfFermionSite(cons_N=cons_N, cons_Sz=cons_Sz)
+        print(sorted(site.opnames))
+        print(site.state_labels)
+        return site
+
+    def init_lattice(self, params):
+        L = params.get("L", 1)
+        norb = params.get("norb", 4)
+        site = self.init_sites(params)
+        lat = MolecularChain(L, norb, site, basis=[[norb, 0], [0, 1]])
+        return lat
+
+    def init_terms(self, params):
+        dx0 = np.array([0, 0])
+        norb = params.get("norb", 4)
+        one_body_tensor = params.get("one_body_tensor", np.zeros((norb, norb)))
+        two_body_tensor = params.get(
+            "two_body_tensor", np.zeros((norb, norb, norb, norb))
+        )
+        constant = params.get("constant", 0)
+
+        for p in range(norb):
+            for q in range(norb):
+                h1 = one_body_tensor[p, q]
+                if p == q:
+                    self.add_onsite(h1, p, "Nu")
+                    self.add_onsite(h1, p, "Nd")
+                    self.add_onsite(constant / norb, p, "Id")
+                else:
+                    self.add_coupling(h1, p, "Cdu", q, "Cu", dx0)
+                    self.add_coupling(h1, p, "Cdd", q, "Cd", dx0)
+
+                for r in range(norb):
+                    for s in range(norb):
+                        h2 = two_body_tensor[p, q, r, s]
+                        if p == q == r == s:
+                            self.add_onsite(h2 / 2, p, "Nu")
+                            self.add_onsite(-h2 / 2, p, "Nu Nu")
+                            self.add_onsite(h2 / 2, p, "Nu")
+                            self.add_onsite(-h2 / 2, p, "Cdu Cd Cdd Cu")
+                            self.add_onsite(h2 / 2, p, "Nd")
+                            self.add_onsite(-h2 / 2, p, "Cdd Cu Cdu Cd")
+                            self.add_onsite(h2 / 2, p, "Nd")
+                            self.add_onsite(-h2 / 2, p, "Nd Nd")
+                        else:
+                            self.add_multi_coupling(
+                                h2 / 2,
+                                [
+                                    ("Cdu", dx0, p),
+                                    ("Cdu", dx0, r),
+                                    ("Cu", dx0, s),
+                                    ("Cu", dx0, q),
+                                ],
+                            )
+                            self.add_multi_coupling(
+                                h2 / 2,
+                                [
+                                    ("Cdu", dx0, p),
+                                    ("Cdd", dx0, r),
+                                    ("Cd", dx0, s),
+                                    ("Cu", dx0, q),
+                                ],
+                            )
+                            self.add_multi_coupling(
+                                h2 / 2,
+                                [
+                                    ("Cdd", dx0, p),
+                                    ("Cdu", dx0, r),
+                                    ("Cu", dx0, s),
+                                    ("Cd", dx0, q),
+                                ],
+                            )
+                            self.add_multi_coupling(
+                                h2 / 2,
+                                [
+                                    ("Cdd", dx0, p),
+                                    ("Cdd", dx0, r),
+                                    ("Cd", dx0, s),
+                                    ("Cd", dx0, q),
+                                ],
+                            )
--- a/python/ffsim/tenpy/util.py
+++ b/python/ffsim/tenpy/util.py
@@ -1,0 +1,46 @@
+from tenpy.networks.mps import MPS
+from tenpy.networks.site import SpinHalfFermionSite
+
+import ffsim
+
+
+def product_state_to_mps(norb, nelec, idx):
+    r"""Return the product state as an MPS.
+
+    Return type:
+        `TeNPy MPS <https://tenpy.readthedocs.io/en/latest/reference/tenpy.networks.mps.MPS.html#tenpy.networks.mps.MPS>`__
+
+    Returns:
+        The product state as an MPS.
+    """
+
+    dim = ffsim.dim(norb, nelec)
+
+    strings = ffsim.addresses_to_strings(
+        range(dim), norb=norb, nelec=nelec, bitstring_type=ffsim.BitstringType.STRING
+    )
+
+    string = strings[idx]
+    up_sector = list(string[0:norb].replace("1", "u"))
+    down_sector = list(string[norb : 2 * norb].replace("1", "d"))
+    product_state = list(map(lambda x, y: x + y, up_sector, down_sector))
+
+    for i, site in enumerate(product_state):
+        if site == "00":
+            product_state[i] = "empty"
+        elif site == "u0":
+            product_state[i] = "up"
+        elif site == "0d":
+            product_state[i] = "down"
+        elif site == "ud":
+            product_state[i] = "full"
+        else:
+            raise ValueError("undefined site")
+
+    # note that the bit positions increase from right to left
+    product_state = product_state[::-1]
+
+    shfs = SpinHalfFermionSite(cons_N="N", cons_Sz="Sz")
+    psi_mps = MPS.from_product_state([shfs] * norb, product_state)
+
+    return psi_mps
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     "qiskit >= 1.1",
     "scipy",
     "typing-extensions",
+    "physics-tenpy"
 ]
 
 [project.urls]
--- a/python/ffsim/hamiltonians/molecular_hamiltonian.py
+++ b/python/ffsim/hamiltonians/molecular_hamiltonian.py
@@ -25,6 +25,7 @@ from typing_extensions import deprecated
 from ffsim.cistring import gen_linkstr_index
 from ffsim.operators import FermionOperator, cre_a, cre_b, des_a, des_b
 from ffsim.states import dim
+from ffsim.tenpy.hamiltonians import MolecularHamiltonianMPO
 
 
 @dataclasses.dataclass(frozen=True)
@@ -119,6 +120,41 @@ class MolecularHamiltonian:
             two_body_tensor=two_body_tensor_rotated,
             constant=self.constant,
         )
+
+    def to_mpo(self, decimal_places=None):
+        r"""Return the Hamiltonian as an MPO.
+
+        Args:
+            decimal_places: The number of decimal places to which to round the input
+                one-body and two-body tensors. Rounding can sometimes reduce the MPO
+                bond dimension.
+
+        Return type:
+            `TeNPy MPO <https://tenpy.readthedocs.io/en/latest/reference/tenpy.networks.mpo.MPO.html#tenpy.networks.mpo.MPO>`__
+
+        Returns:
+            The Hamiltonian as an MPO.
+        """
+
+        if decimal_places:
+            one_body_tensor = np.round(self.one_body_tensor, decimals=decimal_places)
+            two_body_tensor = np.round(self.two_body_tensor, decimals=decimal_places)
+        else:
+            one_body_tensor = self.one_body_tensor
+            two_body_tensor = self.two_body_tensor
+
+        model_params = dict(
+            cons_N="N",
+            cons_Sz="Sz",
+            L=1,
+            norb=self.norb,
+            one_body_tensor=one_body_tensor,
+            two_body_tensor=two_body_tensor,
+            constant=self.constant,
+        )
+        mpo_model = MolecularHamiltonianMPO(model_params)
+
+        return mpo_model.H_MPO
 
     def _linear_operator_(self, norb: int, nelec: tuple[int, int]) -> LinearOperator:
         """Return a SciPy LinearOperator representing the object."""
--- a/python/ffsim/tenpy/hamiltonians/__init__.py
+++ b/python/ffsim/tenpy/hamiltonians/__init__.py
@@ -1,0 +1,21 @@
+# (C) Copyright IBM 2024.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Classes for converting Hamiltonians to TeNPy MPO objects."""
+
+from ffsim.tenpy.hamiltonians.molecular_hamiltonian import (
+    MolecularChain,
+    MolecularHamiltonianMPO,
+)
+
+__all__ = [
+    "MolecularChain",
+    "MolecularHamiltonianMPO",
+]
--- a/tests/python/hamiltonians/molecular_hamiltonian_test.py
+++ b/tests/python/hamiltonians/molecular_hamiltonian_test.py
@@ -22,6 +22,7 @@ import pytest
 import scipy.sparse.linalg
 
 import ffsim
+from ffsim.tenpy.util import product_state_to_mps
 
 
 def test_linear_operator():
@@ -164,6 +165,46 @@ def test_rotated():
     original_expectation = np.vdot(vec, linop @ vec)
     rotated_expectation = np.vdot(rotated_vec, linop_rotated @ rotated_vec)
     np.testing.assert_allclose(original_expectation, rotated_expectation)
+
+
+@pytest.mark.parametrize(
+    "norb, nelec",
+    [
+        (4, (2, 2)),
+        (4, (1, 2)),
+        (4, (0, 2)),
+        (4, (0, 0)),
+    ],
+)
+def test_to_mpo(norb: int, nelec: tuple[int, int]):
+    """Test MPO conversion."""
+    rng = np.random.default_rng()
+
+    # generate a random molecular Hamiltonian
+    one_body_tensor = ffsim.random.random_hermitian(norb, seed=rng)
+    two_body_tensor = ffsim.random.random_two_body_tensor(norb, seed=rng)
+    constant = rng.standard_normal()
+    mol_hamiltonian = ffsim.MolecularHamiltonian(
+        one_body_tensor, two_body_tensor, constant=constant
+    )
+    linop = ffsim.linear_operator(mol_hamiltonian, norb, nelec)
+
+    # convert molecular Hamiltonian to MPO
+    mol_hamiltonian_mpo = mol_hamiltonian.to_mpo()
+
+    # generate a random product state
+    dim = ffsim.dim(norb, nelec)
+    idx = rng.integers(0, high=dim)
+    product_state = np.zeros(dim)
+    product_state[idx] = 1
+
+    # convert random product state to MPS
+    product_state_mps = product_state_to_mps(norb, nelec, idx)
+
+    # test expectation is preserved
+    original_expectation = np.vdot(product_state, linop @ product_state)
+    mpo_expectation = mol_hamiltonian_mpo.expectation_value_finite(product_state_mps)
+    np.testing.assert_allclose(original_expectation, mpo_expectation)
 
 
 @pytest.mark.filterwarnings("ignore::DeprecationWarning")


bartandrews:
Note that the `mkl` and `mkl-devel` packages (and `cython`) needs to be installed in the environment for TeNPy to be compiled against MKL for faster performance, see https://tenpy.readthedocs.io/en/latest/install/extra.html I'm not sure of the best way to incorporate this. It works fine without it, but power users need to make sure that TeNPy is compiled optimally.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,315 @@
+import numpy as np
```
kevinsung, line 1: Since we are simulating fermions, we should define gates with fermionic names rather than their qubit equivalents. The natural gates for fermions are slightly different, and can be found in [this file](https://github.com/qiskit-community/ffsim/blob/main/python/ffsim/gates/basic_gates.py). If I'm not mistakaen, the approximate equivalents would be
- `xy` -> `apply_givens`
- `phase` -> `apply_num_interaction`
- `cphase1` -> `apply_on_site_interaction`

python/ffsim/tenpy/circuits/lucj_circuit.py
```diff
@@ -0,0 +1,122 @@
...
+    # initialize chi_list
+    chi_list: list[int] = []
+
+    # prepare initial Hartree-Fock state
+    psi = product_state_as_mps(norb, nelec, 0)
+
+    # construct the qiskit circuit
+    qubits = QuantumRegister(2 * norb)
+    circuit = QuantumCircuit(qubits)
+    circuit.append(ffsim.qiskit.UCJOpSpinBalancedJW(lucj_operator), qubits)
```
kevinsung, line 56: There's no need to create a Qiskit quantum circuit. You should apply the UCJ operator directly. You can emulate the logic in the [`_apply_unitary_` method](https://github.com/qiskit-community/ffsim/blob/a51163622e4e65393d967aa2bc4655ed523aabb0/python/ffsim/variational/ucj_spin_balanced.py#L479-L518) of `UCJOpSpinBalanced`. This may entail defining higher level gates for the orbital rotation and diagonal Coulomb evolution.

python/ffsim/tenpy/circuits/lucj_circuit.py
```diff
@@ -0,0 +1,122 @@
...
+    phase,
+    xy,
+)
+from ffsim.tenpy.util import product_state_as_mps
+
+
+def lucj_circuit_as_mps(
+    norb: int,
+    nelec: tuple,
+    lucj_operator: "ffsim.variational.ucj_spin_balanced.UCJOpSpinBalanced",
```
kevinsung, line 23: This variable should be named `ucj_op`. You should be able to import `UCJOpSpinBalanced` and use it directly for the type annotation.

kevinsung (changes requested):
Thank you Bart! I'll begin with some high-level comments that will involve some code restructuring.

bartandrews:
> Thank you Bart! I'll begin with some high-level comments that will involve some code restructuring.

Thank you for reviewing this! :smiley: I will implement these changes asap

python/ffsim/tenpy/circuits/lucj_circuit.py
kevinsung: My reading of that is that you can use strings for situations when you can't import the type you need (e.g. due to circular imports). That should not be the case here though. In any case, in the rest of ffsim I've been importing modules only for type hints, so we might as well be consistent until we have a reason to change it.

python/ffsim/tenpy/circuits/lucj_circuit.py
```diff
@@ -0,0 +1,68 @@
...
+        apply_orbital_rotation(
+            np.conj(ucj_op.orbital_rotations[i]).T, psi, eng, chi_list, norm_tol
+        )
+        apply_diag_coulomb_evolution(
+            ucj_op.diag_coulomb_mats[i], psi, eng, chi_list, norm_tol
+        )
+        apply_orbital_rotation(
+            ucj_op.orbital_rotations[i], psi, eng, chi_list, norm_tol
+        )
+        if ucj_op.final_orbital_rotation is not None:
```
kevinsung, line 63: This is a bug; this if block should be de-indented one level. Before fixing this bug, add a test with `n_reps` larger than 1 and confirm that the test fails.

python/ffsim/tenpy/circuits/lucj_circuit.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # prepare initial Hartree-Fock state
+    psi = product_state_as_mps(norb, nelec, 0)
+
+    # define the TEBD engine
+    eng = TEBDEngine(psi, None, options)
+
+    # construct the LUCJ MPS
+    n_reps = np.shape(ucj_op.orbital_rotations)[0]
+    for i in range(n_reps):
```
kevinsung, line 53: Instead, use `zip` to iterate through `ucj_op.diag_coulomb_mats` and `ucj_op.orbital_rotations`.

python/ffsim/tenpy/circuits/lucj_circuit.py
```diff
@@ -0,0 +1,68 @@
...
+        apply_orbital_rotation(
+            np.conj(ucj_op.orbital_rotations[i]).T, psi, eng, chi_list, norm_tol
+        )
+        apply_diag_coulomb_evolution(
+            ucj_op.diag_coulomb_mats[i], psi, eng, chi_list, norm_tol
+        )
+        apply_orbital_rotation(
+            ucj_op.orbital_rotations[i], psi, eng, chi_list, norm_tol
+        )
```
kevinsung, line 62: The last orbital rotation of each iteration can be merged with the first one of the next, reducing the number of operations applied and hence the running time and truncation error. You just need to keep track of the basis at the end of each iteration. See https://github.com/qiskit-community/ffsim/blob/420381f2655fabfa12ed76f7934a968f9536d1e7/python/ffsim/variational/ucj_spin_balanced.py#L479-L518.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,419 @@
...
```
kevinsung, line 1: Missing copyright header.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+
+    # Givens decomposition
+    givens_list, diag_mat = givens_decomposition(mat)
+
+    # apply the Givens rotation gates
+    for gate in givens_list:
+        theta = np.arccos(gate.c)
+        s = np.conj(gate.s)
+        phi = np.real(1j * np.log(-s / np.sin(theta))) if theta else 0
```
kevinsung, line 352: According to https://github.com/qiskit-community/ffsim/blob/420381f2655fabfa12ed76f7934a968f9536d1e7/python/ffsim/qiskit/gates/orbital_rotation.py#L200

this can simply be computed as
```python
theta = math.acos(c)
phi = cmath.phase(s)
```
is this right?

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,419 @@
...
+        psi.canonical_form_finite()
+
+
+def apply_orbital_rotation(
+    mat: np.ndarray,
+    psi: MPS,
+    eng: TEBDEngine,
+    chi_list: list,
+    norm_tol: float = 1e-5,
+) -> None:
```
kevinsung, line 321: Here and elsewhere, we should try to match the API of the `ffsim.gates` module as much as possible, e.g. https://github.com/qiskit-community/ffsim/blob/420381f2655fabfa12ed76f7934a968f9536d1e7/python/ffsim/gates/orbital_rotation.py#L44-L51.
Note that the state is always the first argument. In this case, that means `psi` should be the first argument. Also, for now let's make all of the arguments beside `psi` and `mat` keyword-only arguments, in case we want to change something later and not have it be a breaking change.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,419 @@
...
+
+    # alpha sector / up spins
+    if spin in [Spin.ALPHA, Spin.ALPHA_AND_BETA]:
+        Cdu = shfs_nosym.get_op("Cdu").to_ndarray()
+        Cu = shfs_nosym.get_op("Cu").to_ndarray()
+        JWu = shfs_nosym.get_op("JWu").to_ndarray()
+        Nu = shfs_nosym.get_op("Nu").to_ndarray()
+        #
+        Ggate_a = (
+            np.kron(sp.linalg.expm(1j * phi * Nu), Id)
```
kevinsung, line 93: Are the exponentiated matrices just rotation matrices with a closed-form expression? If so, using that expression would be preferable to calling `expm`.

kevinsung (changes requested):

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,146 @@
...
+# ruff: noqa: N806
+
+
+class MolecularHamiltonianMPOModel(CouplingMPOModel):
+    """Molecular Hamiltonian."""
+
+    def __init__(self, params):
+        if hasattr(self, "flag"):  # only call __init__ once
+            return
+        self.flag = True
```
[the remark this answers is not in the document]
bartandrews: I was actually meaning to ask you about this, because I am not sure. The `__init__` function is needed for consistency with TeNPy code conventions and it can only be called once (because it adds terms to the Hamiltonian). In fact, if you try to call it twice, TeNPy will error because then you will duplicate all of the terms in the Hamiltonian. Now that I have moved `from_molecular_hamiltonian` to a static method within the class, there is a line within this method where I need to create a `MolecularHamiltonianMPOModel` instance. Therefore, there is now a chance that `__init__` will be called twice. To resolve this, I have added a simple flag to `__init__`, which makes sure that it is only executed once (on the first time through). Please let me know if you have a more elegant solution :pray:

python/ffsim/tenpy/circuits/gates.py
bartandrews: Yes, all of the TeNPy fermionic operators, including `Cd`, `C`, `Nu`, `JW`, etc. are simply 4x4 matrices. Therefore, you need to be careful since `Cd` does not correspond directly to the conventional $c^\dagger$ without the appropriate Jordan-Wigner string. More information on this here: https://tenpy.readthedocs.io/en/stable/intro/JordanWigner.html

python/ffsim/tenpy/circuits/gates.py
bartandrews: Yes, up to a minus sign. If use the definition from https://qiskit-community.github.io/ffsim/api/ffsim.html#ffsim.apply_givens_rotation , then we define $s=-e^{-i\phi} \sin\theta$.

python/ffsim/tenpy/circuits/lucj_circuit.py
bartandrews: Thanks for pointing that out, that's really neat!

python/ffsim/tenpy/circuits/gates.py
bartandrews: I can change this to a closed-form expression, which will make the code faster but less readable. If we decide to do this, I recommend to keep the original expression in the comments, because finding the correct Jordan-Wigner strings for the TeNPy fermionic operators is quite tricky.

kevinsung: Yes, please used closed-form expressions insead of `expm`. You can keep the original expression in comments.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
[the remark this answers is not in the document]
kevinsung: Sorry, I still don't understand. Can you give me a code example that would cause `__init__` to be called twice?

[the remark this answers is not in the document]
bartandrews: You're right, I cannot reproduce this anymore. This issue must have resolved itself during the restructuring somehow. I have removed these lines now.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,460 @@
...
+    else:
+        raise ValueError(
+            "only 1-site and 2-site gates implemented for symmetry basis conversion"
+        )
+
+    P = np.zeros(np.shape(gate))
+    for i, s in enumerate(swap_list):
+        P[i, s] = 1
+
+    gate_sym = P.T @ gate @ P
```
kevinsung, line 56: You don't need to construct the permutation matrix explicitly. Instead, use the inverse permutation,
```python
perm = [10, 8, 2, 11, 14, 0, 9, 3, 12, 6, 15, 1, 4, 13, 7, 5]
```
and do
```python
return gate[perm][:, perm]
```

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,460 @@
...
+        Ggate_a = np.eye(16, dtype=complex)
+        c = math.cos(theta)
+        for i in [1, 3, 4, 6, 9, 11, 12, 14]:
+            Ggate_a[i, i] = c
+        s = -cmath.exp(-1j * phi) * math.sin(theta)
+        Ggate_a[1, 4] = -s
+        Ggate_a[3, 6] = -s
+        Ggate_a[9, 12] = s
+        Ggate_a[11, 14] = s
+        Ggate_a[4, 1] = np.conj(s)
```
kevinsung, line 111: use the built-in method `s.conjugate()` instead of calling the Numpy function.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,460 @@
...
+
+    # on-site
+    U1_npc = npc.Array.from_ndarray(U1, [shfs.leg, shfs.leg.conj()], labels=["p", "p*"])
+    psi.apply_local_op(site, U1_npc)
+
+
+def apply_gate2(
+    psi: MPS,
+    U2: np.ndarray,
+    site: int,
```
kevinsung, line 311: Can we make this function accept both site indices, and raise an error if they are not adjacent? Then, if the site indices are not sorted, we can account for that by applying an appropriate permutation to U2, or something similar? If this is possible, then we can remote the unintuitive `conj` argument from `givens_rotation` above.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,460 @@
...
+    for i, s in enumerate(swap_list):
+        P[i, s] = 1
+
+    gate_sym = P.T @ gate @ P
+
+    return gate_sym
+
+
+def givens_rotation(
+    theta: float, spin: Spin, *, conj: bool = False, phi: float = 0.0
```
kevinsung, line 62: I wonder if we can get rid of this `conj` argument; see my comment on `apply_gate2`

kevinsung (changes requested):

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,460 @@
...
+            eng=eng,
+            chi_list=chi_list,
+            norm_tol=norm_tol,
+        )
+
+    # apply the number interaction gates
+    for i, z in enumerate(diag_mat):
+        theta = float(np.angle(z))
+        apply_gate1(
+            psi, np.exp(1j * theta) * num_interaction(-theta, Spin.ALPHA_AND_BETA), i
```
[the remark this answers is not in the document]
bartandrews: Yes this is a bug, thanks for spotting this. It works without the phase factor

python/ffsim/tenpy/circuits/gates.py
bartandrews: Yes, this is possible. I think that it is as simple as transposing the gate.

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,452 @@
...
+            defined in the
+            `TeNPy documentation <https://tenpy.readthedocs.io/en/latest/reference/tenpy.algorithms.dmrg.DMRGEngine.html#cfg-option-DMRGEngine.norm_tol>`__.
+
+    Returns:
+        None
+    """
+
+    # extract norb
+    assert mat.shape[1] == mat.shape[2]
+    norb = mat.shape[1]
```
kevinsung, line 432:
```suggestion
    norb, _ = mat.shape
```

tests/python/tenpy/lucj_circuit_test.py
```diff
@@ -0,0 +1,119 @@
...
+    params = rng.uniform(-10, 10, size=n_params)
+    lucj_op = ffsim.UCJOpSpinBalanced.from_parameters(
+        params,
+        norb=norb,
+        n_reps=n_reps,
+        interaction_pairs=_interaction_pairs_spin_balanced_(
+            connectivity=connectivity, norb=norb
+        ),
+        with_final_orbital_rotation=True,
+    )
```
kevinsung, line 103: Use [`ffsim.random.random_ucj_op_spin_balanced`](https://github.com/qiskit-community/ffsim/blob/87c1e7628fa59e622fc28e57ee92bd8bc7130994/python/ffsim/random/random.py#L352) instead.

tests/python/tenpy/lucj_circuit_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # generate the corresponding LUCJ circuit
+    qubits = QuantumRegister(2 * norb)
+    circuit = QuantumCircuit(qubits)
+    circuit.append(ffsim.qiskit.PrepareHartreeFockJW(norb, nelec), qubits)
+    circuit.append(ffsim.qiskit.UCJOpSpinBalancedJW(lucj_op), qubits)
+    lucj_state = ffsim.qiskit.final_state_vector(circuit).vec
```
kevinsung, line 110:
```suggestion
    lucj_state = ffsim.hartree_fock(norb, nelec)
    lucj_state = ffsim.apply_unitary(lucj_state, lucj_op, norb, nelec)
```
No need to create a Qiskit circuit here.

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,60 @@
...
+
+from __future__ import annotations
+
+from tenpy.networks.mps import MPS
+from tenpy.networks.site import SpinHalfFermionSite
+
+import ffsim
+
+
+def product_state_as_mps(norb: int, nelec: int | tuple[int, int], idx: int) -> MPS:
```
kevinsung, line 19: Let's refactor this function to have this API:
```python
def bitstring_to_mps(bitstring: tuple[int, int]) -> MPS
```
Here the input is a tuple of the alpha string and the beta string: `(str_a, str_b)`. I think this API is more straightforward.

kevinsung (changes requested):

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,133 @@
...
+from tenpy.networks.site import SpinHalfFermionSite
+
+from ffsim.hamiltonians.molecular_hamiltonian import MolecularHamiltonian
+from ffsim.tenpy.hamiltonians.lattices import MolecularChain
+
+# ignore lowercase variable checks to maintain TeNPy naming conventions
+# ruff: noqa: N806
+
+
+class MolecularHamiltonianMPOModel(CouplingMPOModel):
```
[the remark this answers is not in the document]
bartandrews: Good idea, we can contribute this. TeNPy currently has 17 open pull requests, some of which have been open for years. I will try to fast-track this one, but it might not be worth waiting for it. In the long term, I would also like to add basic functionality, such as the `apply_gate1` and `apply_gate2` to TeNPy, as well as some common gates and circuit tutorials.

python/ffsim/tenpy/circuits/gates.py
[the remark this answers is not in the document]
bartandrews: In order to catch this bug, I have added a test that checks the preservation of the (complex) matrix element $\langle \psi | H | \phi \rangle$, where $|\psi \rangle$ is a random product state, $H$ is a random molecular Hamiltonian, and $\phi = U |\psi \rangle$ is random orbital rotation applied to $|\psi \rangle$.

python/ffsim/tenpy/hamiltonians/lattices.py
```diff
@@ -0,0 +1,31 @@
...
+# that they have been altered from the originals.
+
+import numpy as np
+from tenpy.models.lattice import Lattice
+
+# ignore lowercase argument checks to maintain TeNPy naming conventions
+# ruff: noqa: N803
+
+
+class MolecularChain(Lattice):
```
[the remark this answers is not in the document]
bartandrews: I have removed it

python/ffsim/tenpy/util.py
bartandrews: Sounds good, using this change I don't need `nelec` anymore. Also, if I take `bitstring: tuple[str, str]`, I don't need `norb` either, since I can use the `len()` function

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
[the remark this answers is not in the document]
bartandrews: I have opened a PR here tenpy/tenpy#471

python/ffsim/tenpy/circuits/gates.py
```diff
@@ -0,0 +1,451 @@
...
+    if sites[0] > sites[1]:
+        U2 = U2.T
+
+    # apply NN gate between (site1, site2)
+    U2_npc = npc.Array.from_ndarray(
+        U2, [shfsc, shfsc.conj()], labels=["(p0.p1)", "(p0*.p1*)"]
+    )
+    U2_npc_split = U2_npc.split_legs()
+    eng.update_bond(max(sites), U2_npc_split)
+    chi_list.append(psi.chi)
```
kevinsung, line 339: This line is the first line in the function that seems to use `psi`. Is `psi` used implicitly before this? Is it stored in the engine? Sorry, I am not familiar with TenPY.

python/ffsim/tenpy/circuits/lucj_circuit.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def lucj_circuit_as_mps(
+    norb: int,
+    nelec: int | tuple[int, int],
+    ucj_op: UCJOpSpinBalanced,
+    options: dict,
+    *,
+    norm_tol: float = 1e-5,
+) -> tuple[MPS, list[int]]:
```
kevinsung, line 33: To be more consistent with the rest of ffsim's API, this function should have this API:
```python
def lucj_circuit_as_mps(
    mps: MPS,
    ucj_op: UCJOpSpinBalanced,
    norb: int,
    nelec: int | tuple[int, int],
    options: dict,
    *,
    norm_tol: float = 1e-5,
) -> MPS:
```
I think the output should just be the MPS. If we want to get the bond dimensions, we should figure out a way to extract that from the engine separately. Maybe we can use a callback function.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,143 @@
...
+        dx0 = np.array([0, 0])
+        norb = params.get("norb", 4)
+        one_body_tensor = params.get("one_body_tensor", np.zeros((norb, norb)))
+        two_body_tensor = params.get(
+            "two_body_tensor", np.zeros((norb, norb, norb, norb))
+        )
+        constant = params.get("constant", 0)
+
+        for p in range(norb):
+            for q in range(norb):
```
kevinsung, line 60: Use `itertools.product` to reduce the required levels of indentation.

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,56 @@
...
+    Args:
+        bitstring: The bitstring in the form `(string_a, string_b)`.
+
+    Returns:
+        The product state as an MPS.
+    """
+
+    # extract norb
+    assert len(bitstring[0]) == len(bitstring[1])
+    norb = len(bitstring[0])
```
kevinsung, line 29:
```suggestion
    string_a, string_b = bitstring
    assert len(string_a) = len(string_b)
    norb = len(string_a)
```
Unpacking the strings will improve readability.

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # merge bitstrings
+    up_sector = list(bitstring[0].replace("1", "u"))
+    down_sector = list(bitstring[1].replace("1", "d"))
```
kevinsung, line 33: `list` is unnecessary

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,56 @@
...
+    product_state = list(map(lambda x, y: x + y, up_sector, down_sector))
```
kevinsung, line 34: Prefer list comprehension to `map`, e.g.
```python
product_state = [a + b for a, b in zip(up_sector, down_sector)]
```

kevinsung (changes requested):

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,143 @@
...
+                    self.add_onsite(constant / norb, p, "Id")
+                else:
+                    self.add_coupling(h1, p, "Cdu", q, "Cu", dx0)
+                    self.add_coupling(h1, p, "Cdd", q, "Cd", dx0)
+
+                for r in range(norb):
+                    for s in range(norb):
+                        h2 = two_body_tensor[q, p, s, r]
+                        if p == q == r == s:
+                            self.add_onsite(h2 / 2, p, "Nu")
```
kevinsung, line 74: Nit: Prefer multiplication when it's exact, since it's usually faster than division. I.e. `0.5 * h2`

python/ffsim/tenpy/circuits/lucj_circuit.py
bartandrews: Yes, I think the most important piece of information from `chi_list` is the maximum bond dimension reached during the simulation, and this you can see by simply looking at the bond dimensions of the resulting mps. The added value of `chi_list` is quite niche, since it can tell you specifically at which points in the circuit the bond dimension grows and by how much. I can remove it for now.

python/ffsim/tenpy/circuits/gates.py
bartandrews: Yes, the line `eng.update_bond` implicitly updates `psi`

python/ffsim/tenpy/circuits/lucj_circuit.py
bartandrews: Another option is that I create a `CircuitMPS` class that inherits from `MPS` and has some additional attributes, such as `chi_list`.

python/ffsim/tenpy/circuits/gates.py
kevinsung: So `eng` contains a reference to `psi` then? Does `psi` also contain a reference to `eng`? If either of these is the case, then we only need to pass one of them to the function, right?

python/ffsim/tenpy/circuits/lucj_circuit.py
bartandrews: I think that we want to have something like `chi_list` because it tells us explicitly when the bond dimension is saturated, rather than relying solely on observables, such as energy convergence. But I'm not sure how best to do it...

kevinsung: since `eng` stores a reference to the MPS, can we just pass that as input to this function? See my other comment. Also, btw I updated the first post of this thread, I meant to suggest a new name for this function, `apply_ucj_op_spin_balanced`.

python/ffsim/tenpy/circuits/gates.py
bartandrews: The `TEBDEngine` is designed to run the TEBD algorithm (this was originally imagined to be some lengthy time evolution, not simply the application of a single gate). As inputs, it takes the wavefunction (`psi`), the model, and some options. After the engine has run, `psi` will have been updated in-place. Although the engine knows about `psi`, there is no public-facing attribute to get `psi`. The closest we have is [get_resume_data](https://tenpy.readthedocs.io/en/latest/reference/tenpy.algorithms.tebd.TEBDEngine.html#tenpy.algorithms.tebd.TEBDEngine.get_resume_data), which was originally intended to be a checkpoint when running some lengthy time evolution on a cluster. This function uses the internal attribute `self._resume_psi` from the `Algorithm` class to get `psi`, I think. If you're happy using internal attributes, then `eng` contains a reference to `psi`. However, `psi` does not know about `eng`.

bartandrews: TeNPy was not designed with quantum circuits in mind :sweat_smile:

python/ffsim/tenpy/circuits/lucj_circuit.py
bartandrews: Yes, we can just pass `eng`. We can simply use `eng._resume_psi` to get `psi`, if needed.

tests/python/tenpy/gates/ucj_test.py
```diff
@@ -0,0 +1,101 @@
...
+    # generate the corresponding LUCJ circuit MPS
+    dim = ffsim.dim(norb, nelec)
+    strings_a, strings_b = ffsim.addresses_to_strings(
+        range(dim),
+        norb=norb,
+        nelec=nelec,
+        bitstring_type=ffsim.BitstringType.STRING,
+        concatenate=False,
+    )
+    wavefunction_mps = bitstring_to_mps((strings_a[0], strings_b[0]))
```
kevinsung, line 93:
```suggestion
    n_alpha, n_beta = nelec
    wavefunction_mps = bitstring_to_mps(((1 << n_alpha) - 1, (1 << n_beta) - 1))
```

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def init_sites(self, params):
+        cons_N = params.get("cons_N", "N")
+        cons_Sz = params.get("cons_Sz", "Sz")
+        site = SpinHalfFermionSite(cons_N=cons_N, cons_Sz=cons_Sz)
+        return site
+
+    def init_lattice(self, params):
+        L = params.get("L", 1)
+        norb = params.get("norb", 4)
```
kevinsung, line 37: Why does `norb` default to 4?

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,143 @@
...
+            molecular_hamiltonian: The molecular Hamiltonian.
+
+        Returns:
+            The molecular Hamiltonian as an MPO model.
+        """
+
+        model_params = dict(
+            cons_N="N",
+            cons_Sz="Sz",
+            L=1,
```
kevinsung, line 136: If these parameters can't take any other values, is it necessary to actually have them in the dictionary?

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: No reason, I will remove this

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: In TeNPy, these can take other values, but in ffsim I think that we want everything to be (N,Sz)-conserved, so I can remove these options. After all, N and Sz are both conserved by default in SpinHalfFermionSite

tests/python/tenpy/gates/ucj_test.py
bartandrews: That is much cleaner 🙂

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
kevinsung: Yes, please remove any unnecessary options

tests/python/tenpy/gates/ucj_test.py
bartandrews: Is there also a compact way of writing `bitstring_to_mps((strings_a[idx], strings_b[idx]))`, where `idx` is a random integer ?

bartandrews: More generally, I am wondering about the best API for the function `bitstring_to_mps`. Currently, it takes `tuple[int, int]` and `norb`. However, it could instead take simply `int[str, str]`. I have also noticed that you have a `BitstringType` class with some conversion functions.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Args:
+            molecular_hamiltonian: The molecular Hamiltonian.
+
+        Returns:
+            The molecular Hamiltonian as an MPO model.
+        """
+
+        model_params = dict(
+            norb=molecular_hamiltonian.norb,
```
kevinsung, line 136: This one is unnecessary too, since you can get it as `one_body_tensor.shape[0]`

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class MolecularHamiltonianMPOModel(CouplingMPOModel):
+    """Molecular Hamiltonian."""
+
+    def init_sites(self, params):
+        cons_N = params.get("cons_N", "N", expect_type=str)
+        cons_Sz = params.get("cons_Sz", "Sz", expect_type=str)
+        site = SpinHalfFermionSite(cons_N=cons_N, cons_Sz=cons_Sz)
+        return site
```
kevinsung, line 33:
```suggestion
        return SpinHalfFermionSite()
```
This works, right? It looks like SpinHalfFermionSite has the correct arguments by default.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def init_lattice(self, params):
+        L = params.get("L", 1, expect_type=int)
```
kevinsung, line 36: We don't specify the `L` right? I suppose you can just assume it's one.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,142 @@
...
+        norb = params.get("norb", None, expect_type=int)
+        site = self.init_sites(params)
+        basis = np.array(([norb, 0.0], [0, 1]))
```
kevinsung, line 39: Is there a reason you're mixing ints and floats here?

python/ffsim/tenpy/gates/abstract_gates.py
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        eng: The TEBD Engine.
+        U1: The single-site quantum gate.
+        site: The gate will be applied to `site` on the MPS.
+
+    Returns:
+        None
+    """
+
+    # apply single-site gate
```
kevinsung, line 38: Nit: unnecessary whitespace and comment

python/ffsim/tenpy/gates/abstract_gates.py
```diff
@@ -0,0 +1,81 @@
...
+    *,
+    norm_tol: float = 1e-5,
+) -> None:
+    r"""Apply a two-site gate to an MPS.
+
+    Args:
+        eng: The TEBD Engine.
+        U2: The two-site quantum gate.
+        sites: The gate will be applied to adjacent sites `(site1, site2)` on the MPS.
+        norm_tol: The norm error above which we recanonicalize the MPS.
```
kevinsung, line 57: Could you expand the explanation of this argument to make it more accessible to people who are not too familiar with the technical details of MPS?

python/ffsim/tenpy/gates/basic_gates.py
```diff
@@ -0,0 +1,261 @@
...
+        theta: The rotation angle.
+
+    Returns:
+        The on-site interaction gate in the TeNPy (N, Sz)-symmetry-conserved basis.
+    """
+
+    # # using TeNPy SpinHalfFermionSite(cons_N=None, cons_Sz=None) operators
+    # OSgate = sp.linalg.expm(1j * theta * Nu @ Nd)
+    OSgate = np.eye(4, dtype=complex)
+    OSgate[3, 3] = cmath.exp(1j * theta)
```
kevinsung, line 203:
```suggestion
    OSgate = np.diag([1, 1, 1, cmath.exp(1j * theta)])
```

Similarly for other places where you construct diagonal matrices.

python/ffsim/tenpy/gates/diag_coulomb.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # extract norb
+    norb = eng.get_resume_data()["psi"].L
+
+    # unpack alpha-alpha and alpha-beta matrices
+    mat_aa, mat_ab = mat
+
+    # apply alpha-alpha gates
+    for i, j in itertools.product(range(norb), repeat=2):
+        if j > i and mat_aa[i, j]:
```
kevinsung, line 48: You can use `itertools.combinations` to get the unique pairs, instead of what you're doing here with `itertools.product`.

tests/python/tenpy/hamiltonians/molecular_hamiltonian_test.py
```diff
@@ -0,0 +1,65 @@
...
+    mol_hamiltonian_mpo_model = MolecularHamiltonianMPOModel.from_molecular_hamiltonian(
+        mol_hamiltonian
+    )
+    mol_hamiltonian_mpo = mol_hamiltonian_mpo_model.H_MPO
+
+    # generate a random product state
+    dim = ffsim.dim(norb, nelec)
+    idx = rng.integers(0, high=dim)
+    product_state = np.zeros(dim)
+    product_state[idx] = 1
```
kevinsung, line 48:
```suggestion
    product_state = ffsim.linalg.one_hot(dim, idx)
```

tests/python/tenpy/hamiltonians/molecular_hamiltonian_test.py
```diff
@@ -0,0 +1,65 @@
...
+    strings_a, strings_b = ffsim.addresses_to_strings(
+        range(dim),
+        norb=norb,
+        nelec=nelec,
+        bitstring_type=ffsim.BitstringType.STRING,
+        concatenate=False,
+    )
+    product_state_mps = bitstring_to_mps(
+        (int(strings_a[idx], 2), int(strings_b[idx], 2)), norb
+    )
```
kevinsung, line 60:
```suggestion
    strings_a, strings_b = ffsim.addresses_to_strings(
        [idx],
        norb=norb,
        nelec=nelec,
        bitstring_type=ffsim.BitstringType.STRING,
        concatenate=False,
    )
    product_state_mps = bitstring_to_mps(
        (int(strings_a[0], 2), int(strings_b[0], 2)), norb
    )
```

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,142 @@
...
+                    self.add_onsite(0.5 * h2, p, "Nu")
+                    self.add_onsite(-0.5 * h2, p, "Nu Nu")
+                    self.add_onsite(0.5 * h2, p, "Nu")
+                    self.add_onsite(-0.5 * h2, p, "Cdu Cd Cdd Cu")
+                    self.add_onsite(0.5 * h2, p, "Nd")
+                    self.add_onsite(-0.5 * h2, p, "Cdd Cu Cdu Cd")
+                    self.add_onsite(0.5 * h2, p, "Nd")
+                    self.add_onsite(-0.5 * h2, p, "Nd Nd")
+                else:
+                    self.add_multi_coupling(
```
kevinsung, line 85: I feel that this looping logic can be made a bit clearer by performing the
```
if p == q:
```
and
```
if p == q == r == s:
```
blocks in a separate loop, preceding the main loop. I think doing the one- and two-body tensors in separate loops will also improve the readability.

kevinsung (changes requested):

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: Yes, I can assume it's one

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: Yes, this works with default arguments

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: Agree that this would improve readability. However, wouldn't both of these suggestions introduce more loops and therefore make the code slower?

kevinsung: The total number of items being looped over is the same, so I expect the difference to be negligible. Besides, one of the reasons I ask for this change is that I think it will make it simpler to edit the logic later to reduce redundancy in the loop by taking account of symmetry. For example, since the one-body tensor is Hermitian, you only need to loop over its upper triangle. The two-body tensor also satisfies symmetries as demonstrated in [this test](https://github.com/qiskit-community/ffsim/blob/e509eeca3c1e89e51977d5792f4dbd780299fcda/tests/python/random_test.py#L77-L85)

bartandrews: I have split up the loops, but did not manage to finish simplifying the `two_body_tensor` loop part yet.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,136 @@
...
+            self.add_onsite(h2, p, "Ntot")
+            self.add_onsite(-0.5 * h2, p, "Nu Nu")
+            self.add_onsite(-0.5 * h2, p, "Cdu Cd Cdd Cu")
+            self.add_onsite(-0.5 * h2, p, "Cdd Cu Cdu Cd")
+            self.add_onsite(-0.5 * h2, p, "Nd Nd")
+            self.add_onsite(constant / norb, p, "Id")
+
+        for p, q in itertools.combinations(range(norb), 2):
+            self.add_coupling(
+                one_body_tensor[q, p], p, "Cdu", q, "Cu", dx0, plus_hc=True
```
kevinsung, line 71: Nit: Use the index in the original order when it makes sense, e.g. `[p, q]` rather than `[q, p]`

python/ffsim/testing/testing.py
```diff
@@ -153,3 +153,21 @@ def assert_allclose_up_to_global_phase(
         err_msg=err_msg,
         verbose=verbose,
     )
+
+
+def interaction_pairs_spin_balanced(
```
kevinsung, line 158: I've added this function to `variational.util`, see https://github.com/qiskit-community/ffsim/pull/344

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,149 @@
...
+            for q, r in itertools.combinations_with_replacement(range(norb), 2):
+                if not p == q == r == s:
+                    indices = [(p, q, r, s)]
+                    if p < s:
+                        indices.append((s, q, r, p))
+                    if q < r:
+                        indices.append((p, r, q, s))
+
+                    for idx, (i, j, k, ell) in enumerate(indices):
+                        flag_hc = True if not idx and i < ell and j < k else False
```
kevinsung, line 87: This logic is somewhat convoluted, would it be simplified if we handled the case of two unique indices in the preceding
```
for p, q in itertools.combinations(range(norb), 2):
```
loop? In that loop you currently only handle one-body terms.

tests/python/tenpy/hamiltonians/molecular_hamiltonian_test.py
```diff
@@ -0,0 +1,64 @@
...
+        nelec=nelec,
+        bitstring_type=ffsim.BitstringType.STRING,
+        concatenate=False,
+    )
+    product_state_mps = bitstring_to_mps(
+        (int(strings_a[0], 2), int(strings_b[0], 2)), norb
+    )
+
+    # test expectation is preserved
+    original_expectation = np.vdot(product_state, hamiltonian @ product_state)
```
kevinsung, line 62: For a more thorough test, let's generate two different product states and put one on each side, so we're not just testing the diagonal elements of the Hamiltonian. If it's not too expensive, we can even consider testing every single matrix element.

kevinsung (changes requested):

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: One possibility that could make things more readable, is if we do not make use of the `plus_hc` keyword argument of the `add_multi_coupling` method. In this case, I could simply add an additional `if` statement:

```
if p<s and q<r:
    indices.append((s, r, q, p))
```

kevinsung: I still like my idea of handling some of the two-body terms in the
```python
for p, q in itertools.combinations(range(norb), 2):
```
loop, since that loop currently only handles one-body terms.

tests/python/tenpy/hamiltonians/molecular_hamiltonian_test.py
bartandrews: I have extended the test to check every single matrix element. (I reduced the `norb` parameter from 4 to 2, so that the test runs in a reasonable time.)

tests/python/tenpy/hamiltonians/molecular_hamiltonian_test.py
```diff
@@ -0,0 +1,89 @@
...
+            abs(mpo_expectation.real),
+            rtol=1e-05,
+            atol=1e-08,
+        )
+        np.testing.assert_allclose(
+            abs(original_expectation.imag),
+            abs(mpo_expectation.imag),
+            rtol=1e-05,
+            atol=1e-08,
+        )
```
kevinsung, line 89: Shouldn't the expectation values be the same? Why do you only compare the absolute values of the real and imaginary parts?

bartandrews: Yes, I also think that the expectation values should be the same. The diagonal matrix elements are exactly the same. For the off-diagonal matrix elements, sometimes there is a minus sign for either the real or imaginary part. I tried tinkering with the Hamiltonian to fix this, however I really don't understand why this is... :confused: If you have any ideas, please do let me know.

tests/python/tenpy/gates/basic_gates_test.py
```diff
@@ -0,0 +1,288 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Tests for the TeNPy basic gates."""
```
kevinsung, line 11: I see that in the gate tests, you do the following:
1. Apply the gate to a product state as a state vector
2. Apply the gate to a product state as an MPS
3. Check that the expectation value on a molecular Hamiltonian matches in both cases.

Is it possible to retrieve the state vector amplitudes from the MPS? If so, then we don't need to introduce a Hamiltonian. We can just directly check that amplitudes stored by the MPS match the exact ones.

Another question is whether we can apply the evolution not to a product state, but to a random state instead. The basic gates often have trivial action on a product state. For example, applying a Givens rotation between two occupied orbitals doesn't do anything.

bartandrews: It is possible to convert an MPS to statevector using the [mps_to_full](https://tenpy.readthedocs.io/en/latest/reference/tenpy.algorithms.exact_diag.ExactDiag.html#tenpy.algorithms.exact_diag.ExactDiag.mps_to_full) function. Yes, I agree that we should remove the MolecularHamiltonian from these gate tests, and keep the tests for the molecular Hamiltonian separate.

Generating a completely random MPS is non-trivial, since a generic random state from Hilbert space will have volume law entanglement, whereas MPS are good at representing states with area law entanglement. I presume that you mean a completely random state, and not just a random product state? One solution is to start with a random product state and then perform a TEBD [RandomUnitaryEvolution](https://tenpy.readthedocs.io/en/latest/reference/tenpy.algorithms.tebd.RandomUnitaryEvolution.html#tenpy.algorithms.tebd.RandomUnitaryEvolution) to introduce some short-range entanglement. This will be not result in a completely random state but it will be entangled to as much as the computational cost / bond dimension will allow.

kevinsung: A uniformly random state would be ideal, but perhaps unnecessary. We should be able to represent such a state for small test sizes though, right? You can generate a uniformly random state vector using `ffsim.random.random_state_vector`. Is it possible to convert such a state vector into an MPS using Tenpy? Otherwise, I think your idea of generating a random evolution should be fine.

bartandrews: Yes, for small system sizes you may be able to convert a completely random statevector into an MPS using the [full_to_mps](https://tenpy.readthedocs.io/en/latest/reference/tenpy.algorithms.exact_diag.ExactDiag.html#tenpy.algorithms.exact_diag.ExactDiag.full_to_mps) function. However, the bond dimension will blow up quite quickly. I will try in the first instance and if it's not tractable, I will try the random unitary evolution approach.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,118 @@
...
+    """Molecular Hamiltonian."""
+
+    def init_sites(self, params) -> SpinHalfFermionSite:
+        """Initialize sites."""
+        return SpinHalfFermionSite()
+
+    def init_lattice(self, params) -> Lattice:
+        """Initialize lattice."""
+        one_body_tensor = params.get("one_body_tensor", None, expect_type="array")
+        norb = one_body_tensor.shape[0]
```
kevinsung, line 38: Can `one_body_tensor` end up being `None` after line 37? If so, then line 38 will raise an error.

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,118 @@
...
+            positions=pos,
+        )
+        return lat
+
+    def init_terms(self, params) -> None:
+        """Initialize terms."""
+        one_body_tensor = params.get("one_body_tensor", None, expect_type="array")
+        two_body_tensor = params.get("two_body_tensor", None, expect_type="array")
+        constant = params.get("constant", 0, expect_type="real")
+        norb = one_body_tensor.shape[0]
```
kevinsung, line 55: Same question as above, is it possible that `one_body_tensor` is None in line 55?

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
bartandrews: Yes, that's right. In a previous iteration of the code I passed `norb` as an argument to `MolecularHamiltonianMPOModel`, and then I could initialize `one_body_tensor` as `np.zeros((norb, norb))` by default. Now that I do not have `norb` a priori, I decided to pass `None` as the default value. I will catch this error.

kevinsung: Actually, I think the `one_body_tensor` should always be present, and it should not be allowed to be `None`. Is that compatible with TeNPy's programming model?

bartandrews: The `CouplingMPOModel` in TeNPy takes a dictionary `model_params` containing "all the model parameters" (quote from docstring). For our subclass "MolecularHamiltonianMPOModel", I could pass the `one_body_tensor` argument separate to `model_params` -- that way it would be a required argument of the subclass instance. Otherwise, if we keep it in the `model_params`, the only way I can think of is by catching this exception with an assert/if statement to explicitly check that it's not zero or None.

kevinsung: Oh, I see. Let's stick with how TeNPy does things, and just use `model_params`. Is there any precedent in TeNPy for validating that `model_params` must contain certain keys? Or does it always fall back to some default behavior, as you've done here setting `one_body_tensor` to the zeros matrix? I'm thinking that if we do some validation in `__init__` then that would simplify the rest of the code, which can just assume the presence of the keys. But since I ultimately want this class to go in TeNPy, we should just follow how TeNPy does things.

bartandrews: Yes agreed, I have added a validation in an `__init__` method. This checks for existence and also array type, as in TeNPy's `expect_type=` argument of `Config.get`. The `__init__` method also sets these required parameters as instance variables, which means we no longer duplicate code to fetch `one_body_tensor` (in `init_lattice` and `init_terms`). Since there is no in-built function for this purpose, I think this is the most TeNPy-compatible way to achieve this.

tests/python/tenpy/hamiltonians/molecular_hamiltonian_test.py
bartandrews: For future reference, the sign issue has now been fixed. It was due to different conventions for ordering the fermionic modes: TeNPy uses ⇅⇅⇅⇅, whereas ffsim uses ⇊⇊⇈⇈.

tests/python/tenpy/gates/basic_gates_test.py
bartandrews: Another thought on this, currently we have a `bitstring_to_mps` utility function that can convert a product state into an MPS. If we have an arbitrary state in ffsim, which is some superposition of bitstrings, perhaps we can construct the corresponding MPS by reading off the amplitudes and then using the [MPS.add](https://tenpy.readthedocs.io/en/latest/reference/tenpy.networks.mps.MPS.html#tenpy.networks.mps.MPS.add) function? I am imagining something like a `state_to_mps` function, which is higher level than `bitstring_to_mps`. This may be easier than trying to decipher TeNPy's internal functions.

kevinsung: Sure, that sounds fine with me. Is the `full_to_mps` function not suitable for this task?

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,66 @@
...
+        The bitstring as an MPS.
+    """
+
+    # unpack bitstrings
+    int_a, int_b = bitstring
+    string_a = format(int_a, f"0{norb}b")
+    string_b = format(int_b, f"0{norb}b")
+
+    # relabel using TeNPy SpinHalfFermionSite convention
+    product_state, swap_factor = [], 1
```
kevinsung, line 37: Put these on separate lines.

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,66 @@
...
+                swap_factor *= -1
+
+    # construct product state MPS
+    shfs = SpinHalfFermionSite(cons_N="N", cons_Sz="Sz")
+    mps = MPS.from_product_state([shfs] * norb, product_state)
+
+    # map from ffsim to TeNPy ordering
+    minus_identity_npc = npc.Array.from_ndarray(
+        -shfs.get_op("Id").to_ndarray(), [shfs.leg, shfs.leg.conj()], labels=["p", "p*"]
+    )
```
kevinsung, line 62: This should go inside the `if swap_factor == -1:` block.

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,66 @@
...
+            product_state.append("down")
+        else:  # site == ("1", "1"):
+            product_state.append("full")
+
+        if i > 0:
+            if product_state[-1] in ["up", "full"] and product_state[-2] in [
+                "down",
+                "full",
+            ]:
+                swap_factor *= -1
```
kevinsung, line 53: This can be made more pleasant by first defining a dictionary for the label mapping, e.g.
```
{
"00": "empty",
"10": "up",
"01": "down",
"11": "full"
}
```
Then for the sign check, `product_state[-1]` will already be stored in a variable, and we can introduce another variable `previous_site` to store `product_state[-2]`, which gets updated at every iteration of the loop.

tests/python/tenpy/gates/basic_gates_test.py
bartandrews: It may be... I will check what works best

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,63 @@
...
+    # relabel using TeNPy SpinHalfFermionSite convention
+    product_state = []
+    swap_factor = 1
+    occupation = {"00": 0, "10": 1, "01": 2, "11": 3}
+    previous_site_occupation = None
+    for i, site in enumerate(zip(reversed(string_a), reversed(string_b))):
+        site_occupation = occupation["".join(site)]
+        product_state.append(site_occupation)
+
+        if site_occupation in [1, 3] and previous_site_occupation in [2, 3]:
```
kevinsung, line 45: Nit: Write these in binary to make them easier to interpret
```suggestion
        if site_occupation in [0b01, 0b11] and previous_site_occupation in [0b10, 0b11]:
```

tests/python/tenpy/gates/basic_gates_test.py
bartandrews: I have added the `mps_to_statevector` and `statevector_to_mps` functions. They both use TeNPy's in-built functions `full_to_mps` and `mps_to_full`, although most of the hard work was in converting from ffsim to tenpy bases and getting the minus signs correct, due to the different fermion ordering. I have kept around the function `bitstring_to_mps` because it is faster than `statevector_to_mps`, although these could be combined potentially...

bartandrews: I have now integrated the functionality of `bitstring_to_mps` into `statevector_to_mps` (and removed `bitstring_to_mps`). I have checked that the speed of the `statevector_to_mps` is equivalent to the old `bitstring_to_mps` function in these cases.

python/ffsim/tenpy/util.py
```diff
@@ -0,0 +1,255 @@
...
+    # convert TeNPy MPS to ffsim statevector
+    statevector = cast(NDArray[np.complex128], exact_diag.mps_to_full(mps).to_ndarray())
+    statevector = np.multiply(swap_factors_ffsim, statevector[basis_ordering_ffsim])
+
+    return statevector
+
+
+def statevector_to_mps(
+    statevector: NDArray[np.complex128],
+    mpo_model: MPOModel,
```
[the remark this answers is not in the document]
bartandrews: In TeNPy, the MPOModel is needed to define an ExactDiag class instance (https://tenpy.readthedocs.io/en/v1.0.4/reference/tenpy.algorithms.exact_diag.ExactDiag.html). Once we have this, then we can use the in-built methods `mps_to_full` or `full_to_mps`. However, alternatively, we could simply take `norb` and `nelec`, and then define a dummy SpinHalfFermionSite model (e.g. Fermi-Hubbard model) and pass that. I think that should also work because the model is only needed to get the basis information.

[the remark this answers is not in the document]
kevinsung: >However, alternatively, we could simply take norb and nelec, and then define a dummy SpinHalfFermionSite model (e.g. Fermi-Hubbard model) internally and pass that. I think that should also work because the model is only needed to get the basis information, afaik

If this produces the same results then we should do this. No need to have an unnecessary argument.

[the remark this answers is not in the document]
bartandrews: sounds good, I will try this, and will push if it works

python/ffsim/tenpy/hamiltonians/molecular_hamiltonian.py
```diff
@@ -0,0 +1,45 @@
...
+
+from tenpy.models.molecular import MolecularModel
+
+from ffsim.hamiltonians.molecular_hamiltonian import MolecularHamiltonian
+
+# ignore lowercase variable checks to maintain TeNPy naming conventions
+# ruff: noqa: N806
+
+
+class MolecularHamiltonianMPOModel(MolecularModel):  # type: ignore
```
kevinsung, line 23: We don't need this class anymore, let's just use MolecularModel from TeNPy directly.

PR #349, opened 2025-01-08
Implement trace of FermionOperator

### LINKED ISSUE #348: Implement trace for FermionOperator

minhctran:
Fixes #348


tests/python/operators/trace_test.py
```diff
@@ -0,0 +1,23 @@
...
+import time
+import ffsim
+
+def test_trace(norb = 50, nelec = (3,3)):
+    n = norb
+    
+    h = np.matrix(np.random.rand(n,n))
+    h = h + h.H
+    V = np.array([np.matrix(np.random.rand(n,n)),np.matrix(np.random.rand(n,n))])
+    H = ffsim.DiagonalCoulombHamiltonian(h,V)
```
kevinsung, line 11: Just do
```python
H = ffsim.random.random_diagonal_coulomb_hamiltonian(norb, real=True, seed=rng)
```
where `rng` is your random Generator instance. I just added the `real` flag in https://github.com/qiskit-community/ffsim/pull/350 so you'll need to merge main first.

tests/python/operators/trace_test.py
```diff
@@ -0,0 +1,23 @@
...
+    
+    tic = time.time()
```
kevinsung, line 13: Delete all the timing and print statements.

tests/python/operators/trace_test.py
```diff
@@ -0,0 +1,23 @@
...
+    t1 = ffsim.trace(ffsim.fermion_operator(H),norb=norb, nelec=nelec)
+    toc = time.time()
+    print(f"Trace = {t1}. Took {toc-tic}s by converting DiagonalCoulombHamiltonian to FermionOperator")
+    
+    tic = time.time()
+    t2 = ffsim.trace(H, norb=n, nelec=nelec)
+    toc = time.time()
+    print(f"Trace = {t2}. Took {toc-tic}s using DiagonalCoulombHamiltonian._trace()")
+
+    assert np.abs(t1-t2)<1e-3, "The two methods do not match!"
```
kevinsung, line 23:
```suggestion
    np.testing.assert_allclose(t1, t2)
```
This function gives a more helpful error message.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -14,8 +14,12 @@
 
 from typing import Any, Protocol
 
+from ffsim.operators import FermionOperator
+
 import numpy as np
 
+from math import comb
```
kevinsung, line 21: Nit: Normally we just `import math` rather than import specific functions from it. Then in the code, do `math.comb`.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +52,84 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
         "The object did not have a _trace_ method that returned the trace "
         "or a _diag_ method that returned its diagonal entries."
     )
+
+def _trace_fermion_operator(
+    A: FermionOperator, norb: int, nelec: tuple[int, int]
+):
+    n_elec_alpha, n_elec_beta = nelec
+    if len(A)>1:
```
kevinsung, line 60: The if statement and use of recursion is unnecessary here. Instead, define a helper function that computes the trace for a single term. The contents of that function would basically be the else case here. Then in the main function, simply iterate through the terms, calling the helper function to compute the trace of each term.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +52,84 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+        for op in A:
+            B = FermionOperator({op:A[op]})
+            trace+= _trace_fermion_operator(B, norb, nelec)
+    else:
+        trace = 0
+        op = list(A.keys())[0]
+        coeff = A[op]
+        alpha_indices = []
+        beta_indices = []
+        for a in op:
```
kevinsung, line 71: Unpacking the tuple will make this more readable:
```python
for _, spin, orb in op:
```

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +52,84 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+            if not a[1]:
+                alpha_indices.append(a[2])
+            else:
+                beta_indices.append(a[2])
+        alpha_indices = list(set(alpha_indices))
+        beta_indices = list(set(beta_indices))
+        combined_indices = [(i,False) for i in alpha_indices]  +[(i,True) for i in beta_indices]
+        
+        alpha_mapping = dict([(j,i) for i, j in enumerate(alpha_indices)])
+        beta_mapping = dict([(j,i) for i, j in enumerate(beta_indices)])
```
kevinsung, line 81: These variables are not used.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +52,84 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+        # loop over the support of the operator
+        # assume that each site is either 0 or 1 at the beginning
+        # track the state of the site through the application of the operator
+        # if the state exceed 1 or goes below 0, the state is not physical and the trace must be 0
+        for i, spin in combined_indices:
+            initial_zero = 0
+            initial_one = 1
+            is_zero = True
+            is_one = True
+            for a in reversed(op):
```
kevinsung, line 99: As above, unpacking the tuple will make this more readable:
```python
for action, spin, orb in reversed(op)
```

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +52,84 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+                if a[1] == spin and a[2] == i:
+                    change = a[0]*2-1
+                    initial_zero += change
+                    initial_one += change
+                    if initial_zero < 0 or initial_zero > 1:
+                        is_zero = False
+                    if initial_one < 0 or initial_one > 1:
+                        is_one = False
+                if not is_zero and not is_one:
+                    print((is_zero,is_one))
```
kevinsung, line 109: No print statements please. If you want to log stuff you can use the [logging module](https://docs.python.org/3/library/logging.html).

kevinsung (changes requested):

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+
+    return trace
+
+
+def _trace_fermion_operator(ferm: FermionOperator, norb: int, nelec: tuple[int, int]):
+    n_elec_alpha, n_elec_beta = nelec
+    trace = 0j
+    for op in ferm:
+        single_term = FermionOperator({op: ferm[op]})
+        trace += _trace_fermion_operator_single(single_term, norb, nelec)
```
kevinsung, line 135: Instead of taking a FermionOperator as input, the helper function should just take the term. Then, the main function becomes simply
```python
sum(coeff * _trace_fermion_term(term, norb, nelec) for term, coeff in op.items())
```

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+            break
+
+    if not no_configuration:
+        # the trace is nontrival and is a product of
+        # coeff, and
+        # binom(#orbs not in the support of A, #elec on these orbs)
+        # for each spin species
+
+        trace = coeff * math.comb(norb - len(alpha_indices), n_elec_alpha - eta_alpha)
+        trace = trace * math.comb(norb - len(beta_indices), n_elec_beta - eta_beta)
```
kevinsung, line 125: This should be written as a single statement. If the line is too long, Ruff should format it automatically, I think.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+
+
+def _trace_fermion_operator_single(
+    ferm: FermionOperator, norb: int, nelec: tuple[int, int]
+):
+    n_elec_alpha, n_elec_beta = nelec
+    op = list(ferm.keys())[0]
+    coeff = ferm[op]
+    alpha_indices = []
+    beta_indices = []
```
kevinsung, line 63: Just make these sets to begin with, and `add` to them instead of `append`ing.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    for _, spin, orb in op:
+        if not spin:
+            alpha_indices.append(orb)
+        else:
+            beta_indices.append(orb)
+    alpha_indices = list(set(alpha_indices))
+    beta_indices = list(set(beta_indices))
```
kevinsung, line 70: Converting to `list` is unnecessary, right?

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    combined_indices = [(i, False) for i in alpha_indices] + [
+        (i, True) for i in beta_indices
+    ]
+
+    no_configuration = False
```
kevinsung, line 75: It looks like if this is ever set to True, then the trace is zero. Instead of creating this flag and checking it later, just immediately return 0 in those cases.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    for _, spin, orb in op:
+        if not spin:
+            alpha_indices.append(orb)
+        else:
+            beta_indices.append(orb)
+    alpha_indices = list(set(alpha_indices))
+    beta_indices = list(set(beta_indices))
+    combined_indices = [(i, False) for i in alpha_indices] + [
+        (i, True) for i in beta_indices
+    ]
```
kevinsung, line 73: Instead of creating separate alpha and beta sets and then combining them, just create a combined set from the beginning which stores `(orb, spin)` pairs. Since you need the size of the alpha and beta supports later, you can compute that at the same time if you want.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,87 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    # track the state of the site through the application of the operator
+    # if the state exceed 1 or goes below 0,
+    # the state is not physical and the trace must be 0
+    for i, spin in combined_indices:
+        initial_zero = 0
+        initial_one = 1
+        is_zero = True
+        is_one = True
+        for action, aspin, orb in reversed(op):
+            if aspin == spin and orb == i:
```
kevinsung, line 90: I would do
```
if (aspin, orb) != (spin, i):
    continue
```
Then, the next block can be de-indented one level.

kevinsung (changes requested):

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
         "The object did not have a _trace_ method that returned the trace "
         "or a _diag_ method that returned its diagonal entries."
     )
+
+
+def _trace_term(
+    op: tuple[tuple[bool, bool, int], ...], norb: int, nelec: tuple[int, int]
+):
+    n_elec_alpha, n_elec_beta = nelec
```
kevinsung, line 59: For consistency with the rest of ffsim, use `n_alpha` and `n_beta` for these variable names.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+
+    combined_indices = set([(orb, spin) for _, spin, orb in op])
```
kevinsung, line 61: The square brackets are unnecessary. Also sorry to be annoying but please keep the ordering of the tuple the same, i.e. `(spin, orb)` instead of `(orb, spin)`. Also, this variable name is pretty opaque, I suggest `spin_orbs` instead.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    norb_alpha = sum([not spin for _, spin in combined_indices])
```
kevinsung, line 62: Unnecessary square brackets.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    norb_beta = sum([spin for _, spin in combined_indices])
```
kevinsung, line 63:
```suggestion
    norb_beta = len(combined_indices) - norb_alpha
```
to avoid iterating through `combined_indices` again.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+
+    eta_alpha = 0
+    eta_beta = 0
```
kevinsung, line 66: Let's choose more descriptive names, how about `nelec_alpha` and `nelec_beta` or something like that?

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    # track the state of the site through the application of the operator
+    # if the state exceed 1 or goes below 0,
+    # the state is not physical and the trace must be 0
+    for i, spin in combined_indices:
+        initial_zero = 0
+        initial_one = 1
+        is_zero = True
+        is_one = True
+        for action, aspin, orb in reversed(op):
+            if (aspin, orb) != (spin, i):
```
kevinsung, line 79: Let's choose a consistent naming scheme for the loop variables so that this comparison becomes more obvious. Something like
```python
if (this_spin, this_orb) != (spin, orb):
```
would be easier to read.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+            if not spin:
+                eta_alpha += 1
+            else:
+                eta_beta += 1
+        # the number of electrons exceed the number of allowed electrons
+        if eta_alpha > n_elec_alpha or eta_beta > n_elec_beta:
+            return 0j
+
+    # the trace is nontrival and is a product of
+    # coeff, and
```
kevinsung, line 107: `coeff` is no longer here

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    # binom(#orbs not in the support of A, #elec on these orbs)
+    # for each spin species
+
+    trace = math.comb(norb - norb_alpha, n_elec_alpha - eta_alpha) * math.comb(
+        norb - norb_beta, n_elec_beta - eta_beta
+    )
+
+    return trace
```
kevinsung, line 115:
```suggestion
    return math.comb(norb - norb_alpha, n_elec_alpha - eta_alpha) * math.comb(
        norb - norb_beta, n_elec_beta - eta_beta
    )
```

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+
+
+def _trace_fermion_operator(ferm: FermionOperator, norb: int, nelec: tuple[int, int]):
+    n_elec_alpha, n_elec_beta = nelec
+    trace = sum([coeff * _trace_term(op, norb, nelec) for op, coeff in ferm.items()])
+
+    return trace
```
kevinsung, line 122:
```suggestion
    return sum(coeff * _trace_term(op, norb, nelec) for op, coeff in ferm.items())
```

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+            initial_zero += change
+            initial_one += change
+            if initial_zero < 0 or initial_zero > 1:
+                is_zero = False
+            if initial_one < 0 or initial_one > 1:
+                is_one = False
+            if not is_zero and not is_one:  # no possible initial state
+                return 0j
+        # if neither the state 0 nor 1 returns to the initial state, the trace must be 0
+        if (is_zero and initial_zero != 0) or (is_one and initial_one != 1):
```
kevinsung, line 92: The comment says "neither ... nor" but the code uses an `or` instead of `and`. So that means at this point, one of `is_zero` or `is_one` must be False, right? If so, you should move the assertion in line 95 to be above these lines.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+
+            change = action * 2 - 1
+            initial_zero += change
+            initial_one += change
+            if initial_zero < 0 or initial_zero > 1:
+                is_zero = False
+            if initial_one < 0 or initial_one > 1:
+                is_one = False
+            if not is_zero and not is_one:  # no possible initial state
+                return 0j
```
kevinsung, line 90: I think the logic is a bit clearer if we de-indent this if condition one level, to be at the same level as the following checks.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,72 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+            return 0j
+        # only one case is possible if the operator has support on site i
+        assert not is_zero or not is_one
+        # count the number of electrons
+        if is_one:
+            if not spin:
+                eta_alpha += 1
+            else:
+                eta_beta += 1
+        # the number of electrons exceed the number of allowed electrons
```
kevinsung, line 102:
```suggestion
        # the number of electrons exceeds the number of allowed electrons
```
Also, I think this comment belongs _inside_ the following if block.

kevinsung (changes requested):

tests/python/operators/trace_test.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_trace():
+    norb = 50
+    nelec = (3, 3)
+
+    rng = np.random.default_rng(12345)
+    ham = ffsim.random.random_diagonal_coulomb_hamiltonian(norb, real=True, seed=rng)
+
+    t1 = ffsim.trace(ffsim.fermion_operator(ham), norb=norb, nelec=nelec)
+
```
kevinsung, line 25: Nit: unnecessary whitespace

python/ffsim/protocols/trace_protocol.py
minhctran: good point

python/ffsim/protocols/trace_protocol.py
minhctran: but then we will have to loop through the rest of op unnecessarily?

kevinsung: True, but we're also saving on checking the condition repeatedly in the cases that it keeps passing. In either case, I expect the performance impact to be pretty small, and the extra clarity might be worth it. I don't feel too strongly though so I'll leave it up to you.

tests/python/operators/trace_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_trace():
+    norb = 50
+    nelec = (3, 3)
+    rng = np.random.default_rng(12345)
+    ham = ffsim.random.random_diagonal_coulomb_hamiltonian(norb, real=True, seed=rng)
+    t1 = ffsim.trace(ffsim.fermion_operator(ham), norb=norb, nelec=nelec)
+    t2 = ffsim.trace(ham, norb=norb, nelec=nelec)
+    np.testing.assert_allclose(t1, t2)
```
kevinsung, line 24: Let's add another test where we test with a MolecularHamiltonian, which you can generate like this:
```python
ham = ffsim.random.random_molecular_hamiltonian(norb, seed=rng, dtype=float)
```
you can do it in this same test or create a new one.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,69 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    # binom(#orbs not in the support of op, #elec on these orbs)
+    # for each spin species
+
+    return math.comb(norb - norb_alpha, n_alpha - nelec_alpha) * math.comb(
+        norb - norb_beta, n_beta - nelec_beta
+    )
+
+
+def _trace_fermion_operator(ferm: FermionOperator, norb: int, nelec: tuple[int, int]):
+    return sum([coeff * _trace_term(op, norb, nelec) for op, coeff in ferm.items()])
```
kevinsung, line 119: Unnecessary square brackets.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,69 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
         "The object did not have a _trace_ method that returned the trace "
         "or a _diag_ method that returned its diagonal entries."
     )
+
+
+def _trace_term(
+    op: tuple[tuple[bool, bool, int], ...], norb: int, nelec: tuple[int, int]
+):
```
kevinsung, line 58: Add a return type annotation. Now that I think of it, I suppose the return type of the `trace` function should probably be `complex`.

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,69 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+    # the trace is nontrival and is a product of
+    # binom(#orbs not in the support of op, #elec on these orbs)
+    # for each spin species
+
+    return math.comb(norb - norb_alpha, n_alpha - nelec_alpha) * math.comb(
+        norb - norb_beta, n_beta - nelec_beta
+    )
+
+
+def _trace_fermion_operator(ferm: FermionOperator, norb: int, nelec: tuple[int, int]):
```
kevinsung, line 118: Add a return type annotation.

kevinsung (changes requested):
Thank you! Just a few more things.

tests/python/operators/trace_test.py
```diff
@@ -0,0 +1,24 @@
+# (C) Copyright IBM 2023.
```
kevinsung, line 1:
```suggestion
# (C) Copyright IBM 2025.
```

tests/python/operators/trace_test.py
kevinsung, line 1: Instead of creating a new test file, add the tests to https://github.com/qiskit-community/ffsim/blob/main/tests/python/operators/fermion_operator_test.py

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,85 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
         "The object did not have a _trace_ method that returned the trace "
         "or a _diag_ method that returned its diagonal entries."
     )
+
+
+def _term_phase(op: tuple[tuple[bool, bool, int], ...]) -> int:
+    phase = 0
+    while len(op) > 0:
```
kevinsung, line 58:
```suggestion
    while op:
```

python/ffsim/protocols/trace_protocol.py
```diff
@@ -45,3 +51,85 @@ def trace(obj: Any, norb: int, nelec: tuple[int, int]) -> float:
...
+                nelec_beta += 1
+        if nelec_alpha > n_alpha or nelec_beta > n_beta:
+            # the number of electrons exceeds the number of allowed electrons
+            return 0j
+
+    # the trace is nontrival and is a product of
+    # binom(#orbs not in the support of op, #elec on these orbs)
+    # for each spin species
+    # and a phase that depends on the ordering between the actions
+
```
kevinsung, line 124: Nit: this particular line of whitespace is unnecessary.

tests/python/operators/fermion_operator_test.py
```diff
@@ -643,3 +643,21 @@ def test_mapping_methods():
         ((ffsim.cre_b(1), ffsim.des_b(2)), -0.5j),
         ((ffsim.cre_b(2), ffsim.des_b(1)), 1 - 0.5j),
     }
+
+
+def test_trace():
+    rng = np.random.default_rng(12345)
+    # compare for random_diagonal_coulomb_hamiltonian
```
kevinsung, line 650:
```suggestion
    rng = np.random.default_rng(12345)
    ham: ffsim.DiagonalCoulombHamiltonian | ffsim.MolecularHamiltonian
    
    # compare for random_diagonal_coulomb_hamiltonian
```

I think this should fix the mypy complaint.

Issue #369, opened 2025-03-10
Add support for more Qiskit gates


kevinsung:
The simulation of Qiskit gates is done in this file: https://github.com/qiskit-community/ffsim/blob/main/python/ffsim/qiskit/sim.py

In principle, any gate that conserves particle number can be simulated. Some gates that come to mind:

- [x] [SwapGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.SwapGate)
- [x] [iSwapGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.iSwapGate)
- [x] [ZGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.ZGate)
- [x] [SGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.SGate)
- [x] [SdgGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.SdgGate)
- [x] [TGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.TGate)
- [x] [TdgGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.TdgGate)
- [x] [CZGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CZGate)
- [x] [GlobalPhaseGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.GlobalPhaseGate)
- [x] [CCZGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CCZGate)
- [x] [CRZGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CRZGate)
- [x] [CSGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CSGate)
- [x] [CSdgGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CSdgGate)
- [x] [DiagonalGate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.DiagonalGate)
- [x] [PermutationGate](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.library.PermutationGate)
- [x] [CSwapGate](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.library.CSwapGate)
[...]

PR #437, opened 2025-08-04
Qiskit gates: Support CSwapGate

q-inho:
Part of #369

## Summary
This PR add support Qiskit’s `CSwapGate`. It implements conditional fermionic swaps on the appropriate orbitals for both spinless and spinful circuits and adds tests accordingly.

### CSwapGate handling
- **Spinless circuits:** Let the control qubit map to spatial orbit $c$ and the two target qubits map to orbitals $i, j$. The controlled swap gate exchanges the occupations of $i$ and $j$ only when the control orbital is occupied. In second-quantumization the action can be written as

$$U_{CSWAP} = (1-n_c)1 + n_c U_{FSWAP}(i,j),$$

where $n_c$ is number operator on orbital $c$ and $U_{FSWAP}(i,j)$ is the fermionic swap gate that exchanges the states of orbital $i$ and $j$ with the proper antisymmetric phase.

- **Spinful circuits:** Each qubit is mapped to a spin orbit. Denote the control spin orbital by ($\sigma_c, c$) and the target spin orbitals by ($\sigma_i, i$) and ($\sigma_j, j$) with $\sigma \in \lbrace \alpha, \beta \rbrace$. The gate acts independently in each spin sector that in the $\alpha$ -sector it swaps orbital $i_\alpha$ and $j_\alpha$ conditioned on the occupation of $c_\alpha$, and similarly for the $\beta$ -sector. So

$$U_{CSWAP} = \prod_{\sigma = \alpha, \beta} \left[ (1 - n_{\sigma, c}) 1 + n_{\sigma,c} U_{FSWAP}(\sigma, i, \sigma, j)\right]$$

If any of the three qubits belong to different spin sectors, only the matching sectors participate in the swap.

### DIFF
--- a/python/ffsim/linalg/util.py
+++ b/python/ffsim/linalg/util.py
@@ -59,20 +59,125 @@ def antihermitian_from_parameters(
     Returns:
         The antihermitian matrix.
     """
-    generator = np.zeros((dim, dim), dtype=float if real else complex)
+    mat = np.zeros((dim, dim), dtype=float if real else complex)
     n_triu = dim * (dim - 1) // 2
     if not real:
         # imaginary part
         rows, cols = np.triu_indices(dim)
         vals = 1j * params[n_triu:]
-        generator[rows, cols] = vals
-        generator[cols, rows] = vals
+        mat[rows, cols] = vals
+        mat[cols, rows] = vals
     # real part
     vals = params[:n_triu]
     rows, cols = np.triu_indices(dim, k=1)
-    generator[rows, cols] += vals
-    generator[cols, rows] -= vals
-    return generator
+    mat[rows, cols] += vals
+    mat[cols, rows] -= vals
+    return mat
+
+
+def antihermitian_from_parameters_jax(
+    params: np.ndarray, dim: int, real: bool = False
+) -> jax.Array:
+    """JAX version of antihermitian_from_parameters."""
+    mat = jnp.zeros((dim, dim), dtype=float if real else complex)
+    n_triu = dim * (dim - 1) // 2
+    if not real:
+        # imaginary part
+        rows, cols = jnp.triu_indices(dim)
+        vals = 1j * params[n_triu:]
+        mat = mat.at[rows, cols].set(vals)
+        mat = mat.at[cols, rows].set(vals)
+    # real part
+    vals = params[:n_triu]
+    rows, cols = jnp.triu_indices(dim, k=1)
+    mat = mat.at[rows, cols].add(vals)
+    # the subtract method is only available in JAX starting with Python 3.10
+    mat = mat.at[cols, rows].add(-vals)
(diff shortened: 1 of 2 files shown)


python/ffsim/qiskit/sim.py
```diff
@@ -686,6 +692,16 @@ def _evolve_state_vector_spinful(
         )
         return states.StateVector(vec=vec, norb=norb, nelec=nelec)
 
+    if isinstance(op, CSwapGate):
+        control, i, j = qubit_indices
+        if (i < norb) != (j < norb):
+            raise ValueError(
+                f"Gate of type '{op.__class__.__name__}' must be applied on orbitals "
+                "of the same spin."
+            )
```
kevinsung:
```suggestion
            raise ValueError(
                f"The target orbitals of '{op.__class__.__name__}' must be of the same spin."
            )
```

python/ffsim/qiskit/sim.py
```diff
@@ -856,6 +872,34 @@ def _apply_iswap(
...
+    *,
+    copy: bool = True,
+) -> np.ndarray:
+    if copy:
+        vec = vec.copy()
+    addresses = np.arange(len(vec))
+    strings = states.addresses_to_strings(addresses, norb=norb, nelec=nelec)
+    strings = np.asarray(strings, dtype=int)
+    i, j = target_orbs
+    c = control
```
kevinsung: No need to rename the variable.

python/ffsim/qiskit/sim.py
```diff
@@ -856,6 +872,34 @@ def _apply_iswap(
...
+    control_bits = (strings >> c) & 1
+    target_i_bits = (strings >> i) & 1
+    target_j_bits = (strings >> j) & 1
+    mask = (control_bits == 1) & (target_i_bits != target_j_bits)
+    swap_mask = (1 << i) | (1 << j)
+    strings[mask] ^= swap_mask
+    new_addresses = states.strings_to_addresses(strings, norb=norb, nelec=nelec)
+    new_vec = np.empty_like(vec)
+    new_vec[new_addresses] = vec
+    return new_vec
```
kevinsung: You can just return
```python
vec[new_addresses]
```
without creating `new_vec`

python/ffsim/qiskit/sim.py
q-inho: The code line
```python
c = control
````
is removed

File: python/ffsim/qiskit/sim.py (hunk 3 above)
Comment (q-inho) [reply python/ffsim/qiskit/sim.py]: Thanks. I modified it to
```python
new_addresses = states.strings_to_addresses(strings, norb=norb, nelec=nelec)
return vec[new_addresses]
```

kevinsung (approved):

Issue #384, opened 2025-04-04
Support greater than 63 orbitals


kevinsung:
ffsim currently cannot support more than 63 orbitals, even with only one or two electrons, due to the way PySCF represents electronic configurations with integers. For example:
```python
import ffsim

vec = ffsim.slater_determinant(64, [0])
```
```
Traceback (most recent call last):
  File "/home/kevinsung/projects/ffsim/scratch/large_norb.py", line 3, in <module>
    vec = ffsim.slater_determinant(64, [0])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kevinsung/projects/ffsim/python/ffsim/states/slater.py", line 96, in slater_determinant
    alpha_index = cistring.str2addr(norb, n_alpha, alpha_string)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kevinsung/.pyenv/versions/ffsim/lib/python3.12/site-packages/pyscf/fci/cistring.py", line 385, in str2addr
    raise NotImplementedError('norb >= 64')
NotImplementedError: norb >= 64
```

This would likely be a significant undertaking.

PR #443, opened 2025-08-07
Qiskit gates: Support PermutationGate

q-inho:
Part of https://github.com/qiskit-community/ffsim/issues/369

## Summary
This PR add support Qiskit’s `PermutationGate` and adds tests accordingly.

### Spinless Case
For a permutation $\pi$ acting on spinless orbitals, the simulator decomposes $\pi$ into fermionic swaps and applies them sequentially.
On a computational‑basis state $|n_0 n_1 \dots n_{m-1} \rangle$ with occupations $n_i \in \lbrace 0 , 1 \rbrace$, the gate implements

$$
\hat P_{\pi}=\left(\prod_{(k,j)\in T_\pi}\hat S_{k,j}\right)
$$

where​ $T_\pi$ is any sequence of transpositions generating $\pi$ and $\hat{S}_{k,j}$ is the fermionic swap exchanging orbitals $k$ and $j$.

### Spinful Case
With spinful orbitals split into $\alpha$ and $\beta$ sectors, the simulator enforces that a permutation never mixes spins.
If $\pi$ separates into permutations $\pi_\alpha$​ and $\pi_\beta$​ acting within the $\alpha$ and $\beta$ sectors, then for a basis state $|n^\alpha; n^\beta\rangle = |n_0^\alpha \dots n_{n-1}^\alpha; n_0^\beta \dots n_{n-1}^\beta\rangle$

$$
\hat P_{\pi} = \left(\prod_{(k,j)\in T_{\pi_alpha}}\hat S_{k,j}^\alpha\right)\left(\prod_{(k,j)\in T_{\pi_\beta}}\hat S_{k,j}^\beta\right)
$$

### DIFF
--- a/python/ffsim/qiskit/sim.py
+++ b/python/ffsim/qiskit/sim.py
@@ -30,6 +30,7 @@ from qiskit.circuit.library import (
     DiagonalGate,
     GlobalPhaseGate,
     Measure,
+    PermutationGate,
     PhaseGate,
     RZGate,
     RZZGate,
@@ -379,6 +380,19 @@ def _evolve_state_vector_spinless(
         vec = _apply_iswap(vec, (i, j), norb=norb, nelec=nelec, copy=False)
         return states.StateVector(vec=vec, norb=norb, nelec=nelec)
 
+    if isinstance(op, PermutationGate):
+        perm = list(op.pattern)
+        qs = qubit_indices.copy()
+        for k in range(len(perm)):
+            while perm[k] != k:
+                j = perm[k]
+                vec = _apply_swap(
+                    vec, (qs[k], qs[j]), norb=norb, nelec=nelec, copy=False
+                )
+                perm[k], perm[j] = perm[j], perm[k]
+                qs[k], qs[j] = qs[j], qs[k]
+        return states.StateVector(vec=vec, norb=norb, nelec=nelec)
+
     if isinstance(op, XXPlusYYGate):
         i, j = qubit_indices
         theta, beta = op.params
@@ -713,6 +727,30 @@ def _evolve_state_vector_spinful(
         vec = _apply_iswap(
             vec, (i % norb, j % norb), norb=norb, nelec=nelec, spin=spin, copy=False
         )
+        return states.StateVector(vec=vec, norb=norb, nelec=nelec)
+
+    if isinstance(op, PermutationGate):
+        perm = list(op.pattern)
+        qs = qubit_indices.copy()
+        for k in range(len(perm)):
+            while perm[k] != k:
+                j = perm[k]
+                if (qs[k] < norb) != (qs[j] < norb):
+                    raise ValueError(
+                        f"Gate of type '{op.__class__.__name__}' must be applied on "
+                        "orbitals of the same spin."
+                    )
+                spin = Spin.ALPHA if qs[k] < norb else Spin.BETA
+                vec = _apply_swap(
+                    vec,
+                    (qs[k] % norb, qs[j] % norb),
+                    norb=norb,
+                    nelec=nelec,
+                    spin=spin,
+                    copy=False,
+                )
+                perm[k], perm[j] = perm[j], perm[k]
+                qs[k], qs[j] = qs[j], qs[k]
         return states.StateVector(vec=vec, norb=norb, nelec=nelec)
 
     if isinstance(op, XXPlusYYGate):


python/ffsim/qiskit/sim.py
```diff
@@ -379,6 +380,19 @@ def _evolve_state_vector_spinless(
         vec = _apply_iswap(vec, (i, j), norb=norb, nelec=nelec, copy=False)
         return states.StateVector(vec=vec, norb=norb, nelec=nelec)
 
+    if isinstance(op, PermutationGate):
+        perm = list(op.pattern)
+        qs = qubit_indices.copy()
+        for k in range(len(perm)):
```
kevinsung: I think this can be improved by factoring out a generator function that yields the swap pairs of the permutation decomposition. Then this can look something like
```python
for i, j in _decompose_permutation(perm):
    vec = _apply_swap(vec, (qs[i], qs[j]), ... )

File: python/ffsim/qiskit/sim.py [hunk 2]
```diff
@@ -379,6 +380,19 @@ def _evolve_state_vector_spinless(
...
+            while perm[k] != k:
+                j = perm[k]
+                vec = _apply_swap(
```
kevinsung: This will introduce many "swap defect" sign fixing operations. I wonder if we can use FSWAPS instead of SWAPS, and then only fix the signs a single time at the end. Or maybe it is not straightforward ...

q-inho: Thank you for pointing this out!
I have to switch the implementation to use FSWAPs for each transposition so the $−1$ on $|11\rangle$ is handled locally. So it can removes those repeated sign corrections.

kevinsung: If it's possible to use FSWAPs instead of SWAPs, then the entire permutation can actually be performed as a single orbital rotation.

q-inho:
PermutationGate now uses one pass occupation relabeling and does not call swap or fswap gates. 
The evolution routes through `_apply_permutation_gate_vec`, with the spinful path performing the spin sector check up front. 
Now it invert the gate pattern to compute the destination qubit for each source, map the state to integer occupation strings, clear the destination bits, copy each source bit into its destination, then convert the permuted strings back to addresses and reorder the amplitudes. 
Previously I decomposed the permutation into transpositions and called `_apply_swap` for each pair, which wrapped fswap with swap defect sign fixes and extra number number phases. 
The new approach computes $\psi' = P \psi$ directly by reindexing the basis, so it removes the repeated defect corrections.

python/ffsim/qiskit/sim.py
```diff
@@ -715,6 +726,29 @@ def _evolve_state_vector_spinful(
         )
         return states.StateVector(vec=vec, norb=norb, nelec=nelec)
 
+    if isinstance(op, PermutationGate):
+        perm = list(op.pattern)
```
kevinsung: I think the conversion to list is unnecessary (?)

python/ffsim/qiskit/sim.py
```diff
@@ -715,6 +726,29 @@ def _evolve_state_vector_spinful(
...
+        qs = qubit_indices
```
kevinsung: No need to rename the variable.

python/ffsim/qiskit/sim.py
```diff
@@ -715,6 +726,29 @@ def _evolve_state_vector_spinful(
...
+        inverse_perm = [0] * len(perm)
+        for out_idx, src_idx in enumerate(perm):
+            inverse_perm[src_idx] = out_idx
```
kevinsung: If `perm` can remain a Numpy array, then the inverse permutation can be computed more simply like this:
```python
        inverse_perm = np.empty_like(perm)
        inverse_perm[perm] = np.arange(len(perm))
```

python/ffsim/qiskit/sim.py
```diff
@@ -715,6 +726,29 @@ def _evolve_state_vector_spinful(
...
+        dests = [qs[inverse_perm[i]] for i in range(len(qs))]
+        for src, dest in zip(qs, dests):
+            if (src < norb) != (dest < norb):
+                raise ValueError(
+                    f"Gate of type '{op.__class__.__name__}' must be applied on "
+                    "orbitals of the same spin."
+                )
```
kevinsung:
```suggestion
        if any(
            (src < norb) != (dest < norb) for src, dest in zip(qubit_indices, dests)
        ):
            raise ValueError(
                f"Gate of type '{op.__class__.__name__}' must be applied on "
                "orbitals of the same spin."
            )
```
removes one level of indentation.

Also, do you mind adding a test that triggers this case?

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +801,44 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
     return indices, remaining_circuit
 
 
+def _apply_permutation_gate_vec(
```
kevinsung: The name `_apply_permutation_gate` would be more consistent with the other function names in this file.

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +801,44 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
...
+    vec: np.ndarray,
+    qubit_indices: Sequence[int],
+    perm: Sequence[int],
+    *,
+    norb: int,
+    nelec: int | tuple[int, int],
+    dests: Sequence[int] | None = None,
+) -> np.ndarray:
+    qs = list(qubit_indices)
```
kevinsung: I think the type conversion is unnecessary (?)

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +801,44 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
...
+        inverse_perm = [0] * len(perm)
+        for out_idx, src_idx in enumerate(perm):
+            inverse_perm[src_idx] = out_idx
+        dests = [qs[inverse_perm[i]] for i in range(len(qs))]
+    else:
+        dests = list(dests)
+
+    dest_mask = 0
+    for dest in dests:
+        dest_mask |= 1 << dest
```
kevinsung:
```suggestion
    dest_mask = sum(1 << dest for dest in dests)
```
seems simpler.

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +801,44 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
...
+    nelec: int | tuple[int, int],
+    dests: Sequence[int] | None = None,
+) -> np.ndarray:
+    qs = list(qubit_indices)
+    if dests is None:
+        perm = list(perm)
+        inverse_perm = [0] * len(perm)
+        for out_idx, src_idx in enumerate(perm):
+            inverse_perm[src_idx] = out_idx
+        dests = [qs[inverse_perm[i]] for i in range(len(qs))]
```
kevinsung: I wonder if there's a good way to de-duplicate the computation of the inverse permutation 🤔

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +801,44 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
...
+
+    dest_mask = 0
+    for dest in dests:
+        dest_mask |= 1 << dest
+
+    strings = np.asarray(
+        states.addresses_to_strings(
+            range(len(vec)), norb=norb, nelec=nelec, bitstring_type=BitstringType.INT
+        ),
+        dtype=object,
```
kevinsung: Why do we need to convert to `object` dtype?

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +801,44 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
...
+    )
+    permuted = strings & ~dest_mask
+    for src, dest in zip(qs, dests):
+        permuted |= ((strings >> src) & 1) << dest
+    indices = states.strings_to_addresses(permuted, norb=norb, nelec=nelec)
+    permuted_vec = np.empty_like(vec)
+    permuted_vec[indices] = vec
+    return permuted_vec
```
kevinsung, line 876: The bit manipulation tricks in this function can be a bit hard to follow, can you add some comments explaining the logic?

python/ffsim/qiskit/sim.py
q-inho: Thank you for your comment. 
Initially the purpose to convert to `object` dtype is that to make bitstrings stay as big-ints.
In `_apply_permutation_gate` we do bit operations that can address bit positions larger than 64 when the total qubit index range exceeds 63. 
If it stays as `unit64`, shifts like (strings >> src) with src ≥ 64, or left-shifts to dest ≥ 64, silently zero out due to 64‑bit width, and masks with 1 << dest get truncated. 
That yields incorrect permutation indices for larger systems.
I will add the detail documented note in this PR and ask for your review for validity of this proposed method including bit manipualtion tricks.

kevinsung: I see. Since the permutation must be applied on orbitals of the same spin, and because we don't support more than 63 spatial orbitals anyway (see https://github.com/qiskit-community/ffsim/issues/384), I wonder if it's possible to do the bit manipulations without ever creating integers greater than 64 bits.

python/ffsim/qiskit/sim.py
q-inho: I eventaully couldn’t find a clean way to deduplicate the inverse permutation computation. 
Two conventional options were: (1) introduce a helper to centralize the repeated `inverse_perm` block, or (2) move spincheck into that helper. 
However, in my opinion both approaches don’t align well with the current ffsim code structure.

python/ffsim/qiskit/sim.py
q-inho: For spinful branch, I splitted alpha/beta occupancy to prevent creating integers greater than 64 bits.

python/ffsim/qiskit/sim.py
q-inho: Added a  `test_permutation_gate_cross_spin_error` to exercise the PermutationGate spin sector guard.
The test prepares a spinful Hartree–Fock state with `norb=1`, `nelec=(1, 0)`, applies `PermutationGate([1, 0])` to swap the alpha/beta qubits, and asserts `final_state_vector` raises the expected “same spin” ValueError. This directly triggers the `(src < norb) != (dest < norb)` condition.
However, I have concern that I created this standlone test function, and wonder whether I have better way to merge this test with other test functions.

q-inho:
This is the bit manipulation routine for propsoed Permutation gate implementation.

Given an $n$-bit occupation string $x \in \lbrace 0,1\rbrace^n$ and subset of bit positions $Q = (q_0, \dots, q_{k-1})$, the gate permutes only those bits according to a permutation $\pi$ of $\lbrace 0, \dots, k-1 \rbrace$:

$$
\begin{equation}
y_{q_t} = x_{q_{\pi(t)}}, \qquad (t=0, \dots, k-1), \qquad y_j = x_j \quad \text{for} \quad j \notin Q
\end{equation}
$$

`_apply_permutation_gate` function follow such bit manipulation step that 
1. Convert the pull rule above into push assignments by defining where each source bit ends up 

$$
d_t = q_{\pi^{-1}(t)}.
$$

2. Build a destination mask

$$
M = \sum^{k-1}_{t=1} 2^{d_t},
$$

then clear those bits in one shots by $x \\& \neg M$.

3. For each source position $q_t$, read the source bit

$$
b_t = \lfloor \frac{x}{2^{q_t}} \rfloor \bmod 2
$$

which is equivalently to

$$
b_t = (x \gg q_t) \\& 1),
$$

shift it to its destination, and accumulate by

$$
y = (x \\& \neg M) \lor \sum^{k-1}_{t=1} b_t 2^{d_t}.
$$

This is responsible for clear then fill to avoids write hazards and handles any pattern of $Q$ as long as destinations are distinct.

For spinful case, we treat the $\alpha$ and $\beta$ $n$-bit halves independely then apply the same mask and shift on each half to get $y^{(\alpha)}$ and $y^{(\beta)}$, then recombine by concatenation

$$
y = y^{(\alpha)} \lor (y^{(\beta)} \ll n).
$$

This make sure it preserve $N_{\alpha}$ and $N_{\beta}$ separately.

I will summarize this explanation into docstring of `_apply_permutation_gate` if I get confirmation to proceed.

python/ffsim/qiskit/sim.py
```diff
@@ -767,6 +800,73 @@ def _extract_x_gates(circuit: QuantumCircuit) -> tuple[list[int], QuantumCircuit
...
+        )
+        alpha_bits = np.asarray(strings_a)
+        beta_bits = np.asarray(strings_b)
+        alpha_pairs = [(src, dest) for src, dest in pairs if src < norb]
+        beta_pairs = [(src - norb, dest - norb) for src, dest in pairs if src >= norb]
+
+        permuted_alpha = _permute_bitstrings(alpha_bits, alpha_pairs)
+        permuted_beta = _permute_bitstrings(beta_bits, beta_pairs)
+
+        combined = permuted_alpha + (permuted_beta << norb)
```
kevinsung: Could this create integers greater than 64 bits? Either way, please add a test for a large number of bits. For example, you could create a circuit with 64 orbitals but only two electrons.

python/ffsim/qiskit/sim.py
kevinsung: I think it's okay to move the spin check into the helper function.

python/ffsim/qiskit/sim.py
q-inho: I missed this part. Thanks. However defining `norb >= 64` raise PySCF's cistring `NotImplementedError`.
I made safer implementation that it rank the permuted alpha and beta strings separately via `strings_to_addresses` and recombine the sector addresses.
Also I added the test for largest bit. By defining `norb=63` and `nelec=(1,1)`, it test with 126 qubits total, split into 63-qubit spin sectors with preparing a state with one lectron in alpha and beta.

PR #446, opened 2025-08-15
change the return shape of double factorization

WanHsuanLin:


python/ffsim/linalg/double_factorized_decomposition.py
```diff
@@ -487,23 +487,25 @@ def double_factorized_t2(
...
+          eigenvector gives rise to 2 terms in the decomposition.
     """
     nocc, _, nvrt, _ = t2_amplitudes.shape
     norb = nocc + nvrt
 
     t2_mat = t2_amplitudes.transpose(0, 2, 1, 3).reshape(nocc * nvrt, nocc * nvrt)
-    outer_eigs, outer_vecs = _truncated_eigh(t2_mat, tol=tol, max_vecs=max_vecs)
+    if max_vecs:
+        n_terms = math.ceil(max_vecs // 2)
+        outer_eigs, outer_vecs = _truncated_eigh(t2_mat, tol=tol, max_vecs=n_terms)
```
kevinsung: You don't need to bother passing `max_vecs` to `_truncated_eigh`. Just let it return all the vectors. You will truncate later anyway. This will simplify the code.

In fact, it's possible that `_truncated_eigh` shouldn't take `max_vecs` as an argument at all. If that's a simple refactor, you can do it here too.

python/ffsim/linalg/double_factorized_decomposition.py
```diff
@@ -590,31 +598,35 @@ def reconstruct_t2_alpha_beta(
...
         nocc_a * nvrt_a, nocc_b * nvrt_b
     )
-    left_vecs, singular_vals, right_vecs = _truncated_svd(
-        t2_mat, tol=tol, max_vecs=max_vecs
-    )
+    if max_vecs:
+        n_terms = math.ceil(max_vecs // 4)
+        left_vecs, singular_vals, right_vecs = _truncated_svd(
+            t2_mat, tol=tol, max_vecs=n_terms
+        )
```
kevinsung: Same comment as above, I think we don't need to bother passing `max_vecs` to `_truncated_svd`

tests/python/linalg/double_factorized_decomposition_test.py
```diff
@@ -428,11 +424,14 @@ def test_double_factorized_t2_amplitudes_random(norb: int, nocc: int):
     diag_coulomb_mats, orbital_rotations = double_factorized_t2(t2)
     reconstructed = reconstruct_t2(diag_coulomb_mats, orbital_rotations, nocc=nocc)
     np.testing.assert_allclose(reconstructed, t2, atol=1e-8)
+    n_reps, _, _ = diag_coulomb_mats.shape
+    even_index = [i for i in range(0, n_reps, 2)]
+    odd_index = [i for i in range(1, n_reps, 2)]
```
kevinsung: Can you just directly use the `range` object? If it needs to be a list, then `list(range(...))` is more idiomatic.

tests/python/linalg/double_factorized_decomposition_test.py
```diff
@@ -492,24 +491,30 @@ def test_double_factorized_t2_alpha_beta_random():
     )
     np.testing.assert_allclose(reconstructed, t2ab, atol=1e-8)
 
+    n_reps = len(diag_coulomb_mats)
+    index_0 = [i for i in range(0, n_reps, 4)]
+    index_1 = [i for i in range(1, n_reps, 4)]
+    index_2 = [i for i in range(2, n_reps, 4)]
+    index_3 = [i for i in range(3, n_reps, 4)]
```
kevinsung: Same comment as above about using the `range` object or idiomatic `list(range(...))`.

python/ffsim/linalg/double_factorized_decomposition.py
```diff
@@ -520,6 +519,12 @@ def double_factorized_t2(
         coeffs[:, :, None, None] * eigs[:, :, :, None] * eigs[:, :, None, :]
     )
 
+    orbital_rotations = orbital_rotations.reshape(-1, norb, norb)
+    diag_coulomb_mats = diag_coulomb_mats.reshape(-1, norb, norb)
+    if max_vecs:
+        orbital_rotations = orbital_rotations[:max_vecs]
+        diag_coulomb_mats = diag_coulomb_mats[:max_vecs]
+
     return diag_coulomb_mats, orbital_rotations
```
kevinsung:
```suggestion
    return diag_coulomb_mats[:max_vecs], orbital_rotations[:max_vecs]
```
This works even if `max_vecs` is `None`.

python/ffsim/linalg/double_factorized_decomposition.py
```diff
@@ -646,6 +648,12 @@ def reconstruct_t2_alpha_beta(
     mats_bb = big_diag_coulomb_mats[:, :, norb:, norb:]
     diag_coulomb_mats = np.stack([mats_aa, mats_ab, mats_bb], axis=2)
 
+    orbital_rotations = orbital_rotations.reshape(-1, 2, norb, norb)
+    diag_coulomb_mats = diag_coulomb_mats.reshape(-1, 3, norb, norb)
+    if max_vecs:
+        orbital_rotations = orbital_rotations[:max_vecs]
+        diag_coulomb_mats = diag_coulomb_mats[:max_vecs]
+
     return diag_coulomb_mats, orbital_rotations
```
kevinsung:
```suggestion
    return diag_coulomb_mats[:max_vecs], orbital_rotations[:max_vecs]
```

kevinsung (commented):
Now that `max_vecs` no longer counts the eigenvectors, we should rename it to `max_terms` or something like that.

python/ffsim/linalg/double_factorized_decomposition.py
kevinsung, line 1: The math equation in the docstring of `double_factorized_t2` needs to be updated.

PR #447, opened 2025-08-15
Automatic LUCJ mapping on heavy-hex

ashsaki:
This PR adds a file under `ffsim.qiskit.transpiler_passes` named `lucj_heavy_hex_preset_pass_manager`. The include a function named `generate_preset_pass_manager_lucj_heavy_hex_with_alpha_betas`, which returns a Qiskit preset pass manager that can map a LUCJ circuit adhering to the heavy-hex freindly zig-zag layout. It also returns a list of allowed alpha-beta interactions which is needed for construction HW-compatible LUCJ ansatz.

Tests pending.


python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
kevinsung, line 1: Need to add copyright header (you can copy it from another file).

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+    generate_preset_pass_manager,
+)
+from qiskit.transpiler.passes import ApplyLayout, VF2PostLayout
+from rustworkx import NoEdgeBetweenNodes, PyGraph
+
+import ffsim
+
+
+def _create_two_linear_chains(num_orbitals: int) -> PyGraph:
+    """In zig-zag layout, there are two linear chains (with connecting qubits
```
kevinsung: Each docstring should begin with a one-line summary (see https://peps.python.org/pep-0257/#multi-line-docstrings)

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+    num_orbitals: int,
+    backend_coupling_graph: PyGraph,
+    alpha_beta_indices: list[tuple[int, int]],
+) -> tuple[PyGraph, list[tuple[int, int]]]:
+    """This function creates the complete zigzag graph that _can be mapped_ to
+    a IBM QPU with heavy-hex connectivity (i.e., the zigzag pattern is an
+    isomorphic sub-graph to the QPU/backend coupling graph). The zigzag pattern
+    includes both linear chains (alpha-alpha/beta-beta interactions) and
+    connecting qubits between the linear chains (alpha-beta interactions).
+    The algorithm works as follows: It starts with an interm graph (`graph_new`)
```
kevinsung: "interim" not "interm" ?

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+
+import ffsim
+
+
+def _create_two_linear_chains(num_orbitals: int) -> PyGraph:
+    """In zig-zag layout, there are two linear chains (with connecting qubits
+    between the chains). This function creates those two linear chains which is
+    a rustworkx.
+
+    PyGraph with two disconnected linear chains. Each chain contains `num_orbitals`
```
kevinsung: Bad line spacing?

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+            in the zigzag pattern. While we want as many connecting (alpha-beta) qubits
+            between the linear (alpha-alpha) chains, we cannot accomodate all due to
+            connectivity constraints of backends. This is the maximum number of
+            connecting qubits the zigzag pattern can have while being backend compliant
+            (i.e., isomorphic subgraph to the backend coupling graph).
+    """
+    isomorphic = False
+    graph = _create_two_linear_chains(num_orbitals=num_orbitals)
+
+    graph_new = copy.deepcopy(graph)  # to avoid not bound warning
```
kevinsung, line 134: Can we use the `copy` method on the PyGraph, rather than `deepcopy`?

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+    isomorphic = False
+    graph = _create_two_linear_chains(num_orbitals=num_orbitals)
+
+    graph_new = copy.deepcopy(graph)  # to avoid not bound warning
+    while not isomorphic:
+        graph_new = copy.deepcopy(graph)
+
+        if not alpha_beta_indices:
+            break
```
kevinsung: This check can be merged into the `while` loop condition, right?

tests/python/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager_test.py
kevinsung, line 1: Need to add copyright header.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+            warnings.warn(
+                f"Backend cannot accomodate alpha_beta_incides {alpha_beta_indices}.\n "
+                f"Removing interaction {alpha_beta_indices[-1]} from the end."
+            )
+            del alpha_beta_indices[-1]
+
+    return graph_new, alpha_beta_indices
+
+
+def _make_backend_cmap_pygraph(backend: BackendV2) -> PyGraph:
```
kevinsung: Given that this function only uses `backend.coupling_map.graph` from the backend, I think it should be refactored so that it simply accepts that graph directly as input.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
```diff
@@ -0,0 +1,298 @@
...
+        raise ValueError(
+            f"Not all qubits in the placeholder `initial_layout` is properly set."
+            f"Negative qubit index in `initial_layout`. "
+            f"intial_layout={initial_layout}"
+        )
+
+    return initial_layout[:-num_allowed_alpha_beta_indices], allowed_alpha_beta_indices
+
+
+def generate_pm_and_interactions_lucj_heavy_hex(
```
kevinsung: Let's rename this function to `generate_lucj_heavy_hex_pass_manager`

kevinsung (changes requested):
You have many references to "IBM" backends, but in principle this should work for any Qiskit backend that has heavy-hex connectivity, so please remove those references.

Does this code also work for square lattice connectivity? If not, which part of the code relies on heavy-hex? This wasn't obvious to me.

tests/python/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager_test.py
```diff
@@ -0,0 +1,112 @@
...
+            requested_alpha_beta_indices=None,
+            optimization_level=3,
+            layout_method="placeholder",
+        )
+
+
+test_data1 = [[(num_orbitals + 1, num_orbitals + 1)], [(num_orbitals, num_orbitals)]]
+
+
+@pytest.mark.parametrize("requested_alpha_beta_indices", test_data1)
```
kevinsung: Please place the data directly into the `parametrize` call, rather than creating a variable for it.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,378 @@
...
+from qiskit_ibm_runtime.models.exceptions import BackendPropertyError
+from rustworkx import NoEdgeBetweenNodes, PyGraph
+
+import ffsim
+
+IBM_TWO_Q_GATES = {"cx", "ecr", "cz", "rzz"}
+VF2_CALL_LIMIT = 30_000_000
+
+
+def _create_two_linear_chains(num_orbitals: int) -> PyGraph:
```
kevinsung: Rename `num_orbitals` to `norb` to match with the rest of ffsim.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
ashsaki: Fixed.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
ashsaki: Merged into while loop condition.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
ashsaki: Now, the function also does node/edge pruning based on error thresholds. So, need the backend to fetch error rates.

tests/python/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager_test.py
ashsaki: Done.

ashsaki:
> You have many references to "IBM" backends, but in principle this should work for any Qiskit backend that has heavy-hex connectivity, so please remove those references.

Changed to Qiskit backend. Let me know if you spot more IBM Backend references.

> Does this code also work for square lattice connectivity? If not, which part of the code relies on heavy-hex? This wasn't obvious to me.

Now, it supports both heavy-hex and grid.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: Added. Note that the copied header says 2024.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
ashsaki: Fixed.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
ashsaki: Text has been changed. No such words as 'interim' or 'interm' now.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: `.copy()` creates a shallow copy according Rustworkx docs. So, sticking to `copy.deepcopy`.

tests/python/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager_test.py
ashsaki: Added.

python/ffsim/qiskit/transpiler_passes/lucj_heavy_hex_preset_pass_manager.py
ashsaki: Renamed to `generate_lucj_pass_manager` as it now supports both heavy-hex and grid.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dev = [
     "nbsphinx",
     "pre-commit",
     "pytest",
+    "qiskit-ibm-runtime",
```
kevinsung: nit: we've been keeping it alphabetical so it should go after `qiskit`

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,443 @@
...
+            f"intial_layout={initial_layout[:-num_allowed_alpha_beta_indices]}"
+        )
+
+    return initial_layout[:-num_allowed_alpha_beta_indices], allowed_alpha_beta_indices
+
+
+def generate_lucj_pass_manager(
+    backend: BackendV2,
+    num_orbitals: int,
+    backend_topology: Literal["heavy-hex", "grid"],
```
kevinsung: Let's rename
- `backend_topology` -> `connectivity`
- `grid` -> `square`

to match https://github.com/qiskit-community/ffsim/blob/b4985aa4f77fba0cec1694b92b434c4aa7473514/python/ffsim/variational/util.py#L57-L75

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,443 @@
+# (C) Copyright IBM 2024.
```
kevinsung: Should be 2026

tests/python/qiskit/transpiler_passes/lucj_preset_pass_manager_test.py
```diff
@@ -0,0 +1,178 @@
+# (C) Copyright IBM 2024.
```
kevinsung: Should be 2026

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,443 @@
...
+        )
+
+    return initial_layout[:-num_allowed_alpha_beta_indices], allowed_alpha_beta_indices
+
+
+def generate_lucj_pass_manager(
+    backend: BackendV2,
+    num_orbitals: int,
+    backend_topology: Literal["heavy-hex", "grid"],
+    requested_alpha_beta_indices: Sequence[tuple[int, int]] | None = None,
```
kevinsung: Let's call it `alpha_beta_interaction_pairs`

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,443 @@
...
+        initial_layout[value] = key
+
+    # as grids do not have ancillae qubits, the trailing items
+    # in the ``initial_layout`` will remain as ``-1``. Therefore,
+    # we are ignoring last several qubits in this check.
+    if -1 in initial_layout[:-num_allowed_alpha_beta_indices]:
+        raise ValueError(
+            f"Not all qubits in the placeholder ``initial_layout`` is properly set."
+            f"Negative qubit index in ``initial_layout``. "
+            f"intial_layout={initial_layout[:-num_allowed_alpha_beta_indices]}"
```
kevinsung: typo in `intial_layout`

tests/python/qiskit/transpiler_passes/lucj_preset_pass_manager_test.py
```diff
@@ -0,0 +1,213 @@
...
+    """Test handling of backend with no noise info."""
+    _, _ = generate_lucj_pass_manager(
+        backend=backend_noise_info_none,
+        num_orbitals=num_orbitals,
+        backend_topology="grid",
+        requested_alpha_beta_indices=None,
+        optimization_level=1,
+    )
+
+    assert True
```
kevinsung: Why is this line needed?

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,454 @@
...
+    requested_alpha_beta_indices: Sequence[tuple[int, int]] | None = None,
+    thresh_two_q: float = 1.0,
+    thresh_meas: float = 0.10,
+    **qiskit_pm_kwargs,
+) -> tuple[StagedPassManager, list[tuple[int, int]]]:
+    """Generate a Qiskit preset pass manager for the LUCJ ansatz.
+
+    Construct a pass manager that maps a local unitary cluster Jastrow (LUCJ)
+    ansatz circuit onto a target backend using layout strategies, as described
+    in Motta et al. (2023) (https://pubs.rsc.org/en/content/articlehtml/2023/sc/d3sc02516k).
```
kevinsung: Please format the references in the same way as https://github.com/qiskit-community/ffsim/blob/1e6ef01618850eda79587382516495a97231d9f6/python/ffsim/states/sample_slater.py#L70-L112

That is, add a references section, refer to it, and make the reference clickable with a hyperlink.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,454 @@
...
+    The layout is optimized for heavy-hex or grid backend topologies, subject to
```
kevinsung: Again, instead of "grid" let's say "square" or "square lattice"

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: Technically it need not to be a square. For example, Nighthawk is not a square as its dimension is $12 \times 10$. So, grid is technically more correct.

ashsaki: But I see the point of matching older references.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: `requested_alpha_beta_indices` is indeed requested by the user, no guarantee that it would be allowed by the Backend. Having the word `requested` is more descriptive, and may be renamed to `requested_alpha_beta_interactions`.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
kevinsung: I think of "square" as referring to the unit cell of the lattice.

pyproject.toml
kevinsung: never mind, we are getting rid of this

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
kevinsung: I think it's fine to omit `requested`. In either case, it is a request from the user that doesn't have to be guaranteed. Of course, we should document this clearly.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: The ref here is much longer than the example. So, used just the author name as hyperlink.
```
`Motta et al. (2023)`_: "Bridging physical intuition and hardware efficiency for correlated electronic states: the local unitary cluster Jastrow ansatz for electronic structure."
```

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: Changed to "square".

tests/python/qiskit/transpiler_passes/lucj_preset_pass_manager_test.py
ashsaki: Made it more explicit. When the backend has no error info, the code still works and returns allowed a-b pairs. Now, the test asserts whether we are getting back a-b pairs.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: Fixed.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: Done.

tests/python/qiskit/transpiler_passes/lucj_preset_pass_manager_test.py
ashsaki: Fixed.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
ashsaki: Fixed.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+
+    return backend_coupling_graph
+
+
+def _get_placeholder_layout_and_allowed_interactions(
+    backend: BackendV2,
+    norb: int,
+    connectivity: Literal["heavy-hex", "square"],
+    alpha_beta_interaction_pairs: list[tuple[int, int]],
+    thresh_two_q: float,
```
kevinsung: let's call it `two_qubit_error_threshold`

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+    thresh_meas: float,
```
kevinsung: let's call it `readout_error_threshold`. The IBM quantum platform uses the term "readout" rather than "measurement".

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+    return initial_layout[:-num_allowed_alpha_beta_indices], allowed_alpha_beta_indices
+
+
+def generate_lucj_pass_manager(
+    backend: BackendV2,
+    norb: int,
+    connectivity: Literal["heavy-hex", "square"],
+    alpha_beta_interaction_pairs: Sequence[tuple[int, int]] | None = None,
+    thresh_two_q: float = 1.0,
+    thresh_meas: float = 0.10,
```
kevinsung: Rename these arguments as specified in my other comments:
- `two_qubit_error_threshold`
- `readout_error_threshold`

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+            are dropped from the end of the list until a valid layout is found.
+            List pairs in descending order of priority to control which are dropped
+            first.
+
+            If ``None``, defaults to pairs as follows:
+
+            - **heavy-hex**: ``[(p, p) for p in range(norb) if p % 4 == 0]``
+            - **square**: ``[(p, p) for p in range(norb)]``
+
+            Default: ``None``.
```
kevinsung: No need for a line that simply states the default value, since the documentation will automatically show it.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+
+    References:
+        - `Motta et al. (2023)`_: "Bridging physical intuition and hardware
+          efficiency for correlated electronic states: the local unitary cluster
+          Jastrow ansatz for electronic structure."
+
+    Args:
+        backend: The target Qiskit backend.
+        norb: Number of spatial orbitals. The ansatz uses
+            ``2 * norb`` qubits plus any ancilla qubits.
```
kevinsung:
```suggestion
        norb: Number of spatial orbitals. The ansatz uses
            ``2 * norb`` qubits, and transpilation may add ancilla qubits.
```

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+
+            If ``None``, defaults to pairs as follows:
+
+            - **heavy-hex**: ``[(p, p) for p in range(norb) if p % 4 == 0]``
+            - **square**: ``[(p, p) for p in range(norb)]``
+
+            Default: ``None``.
+        thresh_two_q: Two-qubit gate error threshold. Edges in the backend
+            coupling graph with error rate ``>= thresh_two_q`` are removed.
+            Default: ``1.0`` (removes only completely faulty edges).
```
kevinsung:
```suggestion
            The default value, ``1.0``, results in only completely faulty edges being removed.
```

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+        thresh_meas: Measurement error threshold. Nodes in the backend
+            coupling graph with measurement error ``>= thresh_meas`` are removed.
+            Default: ``0.10``.
```
kevinsung:
```suggestion
```
No need for a line that simply states the default value, since the documentation will automatically show it.

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+
+    Construct a pass manager that maps a local unitary cluster Jastrow (LUCJ)
+    ansatz circuit onto a target backend using layout strategies, as described in
+    `Motta et al. (2023)`_. The layout is optimized for heavy-hex or square
+    backend topologies, subject to hardware connectivity and error constraints.
+
+    References:
+        - `Motta et al. (2023)`_: "Bridging physical intuition and hardware
+          efficiency for correlated electronic states: the local unitary cluster
+          Jastrow ansatz for electronic structure."
```
kevinsung: I would still like to match the formatting at https://github.com/qiskit-community/ffsim/blob/1e6ef01618850eda79587382516495a97231d9f6/python/ffsim/states/sample_slater.py#L70-L112 so we are consistent throughout ffsim.
- In the main body, just say "the reference below"
- The hyperlink should include the paper title.
- If the formatter complains, add `# noqa: E501` like in this example: https://github.com/qiskit-community/ffsim/blob/d3e9cdd3ddf32b16c6ab262a17582b0097fd2d5c/python/ffsim/qiskit/gates/slater_determinant.py#L129

kevinsung (commented):
Thanks Saki, almost there!

python/ffsim/qiskit/transpiler_passes/lucj_preset_pass_manager.py
```diff
@@ -0,0 +1,461 @@
...
+            target=backend.target, strict_direction=False, call_limit=VF2_CALL_LIMIT
+        )
+    )
+    pm.routing.append(
+        ConditionalController(
+            ApplyLayout(), condition=_custom_apply_post_layout_condition
+        )
+    )
+
+    return pm, allowed_alpha_beta_indices
```
kevinsung: Add an argument to the function called `return_alpha_beta_interaction_pairs=False`. When it's False, just return the `pm`, otherwise return the indices too as you do now. This way, the default behavior matches the behavior of `generate_preset_pass_manager`. Sometimes the user might not care about which interactions were actually allowed, but they can set this argument to True if they want to get them.

ashsaki: Need both `pm` and `pairs_ab` from the function as real backend-compatible circuit needs the implementable `pairs_ab`. Always returning it means one less function argument. One can always do the following if `pais_ab` is not needed.
```python
pm, _ = generate_lucj_pass_manager(<args>)
```

kevinsung: The reason I think we should make it optional is that based on the name of the function, users might expect it to be a drop-in replacement for `generate_preset_pass_manager`, and they will end up writing
```python
pm = generate_lucj_pass_manager(<args>)
```
and be surprised when it doesn't work.

kevinsung: Ok, after reviewing your notebook, I think your design makes sense.

docs/how-to-guides/qiskit-lucj.ipynb
```diff
@@ -17,13 +17,18 @@
...
     "from qiskit.circuit import QuantumCircuit, QuantumRegister\n",
     "from qiskit.providers.fake_provider import GenericBackendV2\n",
-    "from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager\n",
+    "from qiskit.transpiler import CouplingMap\n",
+    "\n",
+    "import ffsim\n",
+    "from ffsim.qiskit import generate_lucj_pass_manager\n",
     "\n",
-    "import ffsim"
+    "warnings.filterwarnings(\"ignore\")"
```
kevinsung: Instead of filtering out the warning about the alpha-beta indices being dropped, let the warning occur and explain it in the text.

docs/how-to-guides/qiskit-lucj.ipynb
```diff
@@ -120,16 +162,22 @@
...
     "\n",
     "transpiled.count_ops()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "371b82fb",
+   "metadata": {},
+   "source": [
+    "As we initially set `pairs_ab=None`, the `generate_lucj_pass_manager()` starts with a default one. For square lattice, it is `[(0, 0), (1, 1), (2, 2), ..., (norb-1, norb-1)]`. However, hardware connectivity may not accommodate all starting pairs, and pairs are dropped from the end of the list until a valid layout is found with a warning. In the example above, last four pairs, from `(15, 15)` to `(12, 12)`, are dropped with a warning message showing the dropped pair.\n",
```
kevinsung:
```suggestion
    "As we initially set `pairs_ab=None`, the `generate_lucj_pass_manager()` starts with a default one. For square lattice, it is `[(0, 0), (1, 1), (2, 2), ..., (norb-1, norb-1)]`. However, if the hardware connectivity does not accommodate all starting pairs, then pairs are dropped from the end of the list until a valid layout is found, and a warning is raised. In the example above, last four pairs, from `(15, 15)` to `(12, 12)`, are dropped with a warning message showing the dropped pairs.\n",
```

docs/how-to-guides/qiskit-lucj.ipynb
```diff
@@ -120,16 +162,22 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "371b82fb",
+   "metadata": {},
+   "source": [
+    "As we initially set `pairs_ab=None`, the `generate_lucj_pass_manager()` starts with a default one. For square lattice, it is `[(0, 0), (1, 1), (2, 2), ..., (norb-1, norb-1)]`. However, if the hardware connectivity does not accommodate all starting pairs, then pairs are dropped from the end of the list until a valid layout is found, and a warning is raised. In the example above, last four pairs, from `(15, 15)` to `(12, 12)`, are dropped with a warning message showing the dropped pairs.\n",
+    "\n",
+    "Note that, a user can use a different `pairs_ab` during the `ucj_op` construction instead of the one provided by the `generate_lucj_pass_manager()`. However, doing so may incur a higher SWAP overhead."
```
kevinsung:
```suggestion
    "Note that you can use a different `pairs_ab` during the `ucj_op` construction instead of the one returned by `generate_lucj_pass_manager`, but doing so may incur a higher SWAP overhead."
```

Issue #408, opened 2025-06-13
Support spinless Jordan-Wigner transformation


kevinsung:
Converting a `MolecularHamiltonianSpinless` to a FermionOperator and then passing it through `ffsim.qiskit.jordan_wigner` yields an operator with double the number of qubits than there should be.

kevinsung:
Sure!

PR #479, opened 2025-10-09
Allow Spinless JW

MIWdlB:
Small update relating to #408


MIWdlB:
Can do, I'll get onto it this week

PR #483, opened 2025-10-20
Add `FermionOperator::to_qubit()` method in Rust for faster JW mapping

### LINKED ISSUE #482: Rust implementation of Jordan Wigner mapper
The current Jordan Wigner mapper in `python/qiskit/jordan_wigner.py` is very slow. 

The JW mapper can be implemented with Rust to get a huge performance boost, especially for quantum chemistry problems or bigger systems.

davidomanovic:
### Summary
This PR adds the new optimized method to the Rust `FermionOperator` class:
```
# Python signature
FermionOperator.to_qubit(norb: int | None = None) -> qiskit.quantum_info.SparsePauliOp
```
- Performs Jordan-Wigner transform with block qubit ordering in a optimized way with a 30-40x speedup from the current Python version `python/qiskit/jordan_wigner.py`. It can also be further optimized by parallelization.

- **Note:** This slightly changes the API, by applying `.to_qubit()` to a `FermionOperator` object. I can also rewrite the python function `python/qiskit/jordan_wigner.py` to do this, to keep the old API `jordan_wigner(FermionOperator: op)` instead. 

### Results
30-40x speedup from original method, HPC friendly. Makes circuit simulations faster.

Example runs from cluster on big systems:
1.  **JW map to qubit Hamiltonian for N2(10e,26o) cc-pvdz**
- Python version `ffsim/python/qiskit/jordan-wigner.py`: 330 seconds
- Rust version `ffsim/src/fermion_operator.rs`: 12 seconds

2.  **JW map to qubit Hamiltonian for N2(10e,26o) 6-31g**
- Python version `ffsim/python/qiskit/jordan-wigner.py`: 46 seconds
- Rust version `ffsim/src/fermion_operator.rs`: 1 second

3.  **Circuit simulation on Qiskit AerSimulator (N2(10e,26o) cc-pvdz)** 

Actually faster with our new qubit Hamiltonian due to better ordering of Pauli strings
```
Simulation time E_rust: 1030.22
Simulation time E_python: 1312.41
R = 1.10, E_rust:-108.9743280336, E_python-108.9743280336
```

### Checklist
- Builds with cargo build
- Passes tox and pytest
- Clippy clean
- Formatted, passes linting...
- Passes workflows

### Issue
Closes #482

### DIFF
--- a/python/ffsim/qiskit/jordan_wigner.py
+++ b/python/ffsim/qiskit/jordan_wigner.py
@@ -20,7 +20,7 @@ from ffsim.operators import FermionOperator
 
 
 def jordan_wigner(op: FermionOperator, norb: int | None = None) -> SparsePauliOp:
-    r"""Jordan-Wigner transformation.
+    r"""Jordan-Wigner transformation as implemented in `src/fermion_operator.rs`.
 
     Transform a fermion operator to a qubit operator using the Jordan-Wigner
     transformation. The Jordan-Wigner transformation maps fermionic annihilation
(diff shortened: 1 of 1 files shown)


kevinsung:
Thank you very much for this! I do prefer to keep the original API, `ffsim.qiskit.jordan_wigner(op)`.

davidomanovic:
I kept the old API by just calling the Rust method inside the `python/qiskit/jordan_wigner.py`, see commit `6232126`.

kevinsung:
> I kept the old API by just calling the Rust method inside the `python/qiskit/jordan_wigner.py`, see commit `88ee968`.

Thanks, but I'd rather not have the `to_qubit` method at all. Can we instead make it a function called `jordan_wigner` and add it to https://github.com/qiskit-community/ffsim/blob/main/python/ffsim/_lib.pyi?

src/fermion_operator.rs
```diff
@@ -337,6 +339,185 @@ impl FermionOperator {
...
+                .into_py(py);
+            let z = Complex64::new(0.0, 0.0).into_py(py);
+            let tup = PyTuple::new_bound(py, &[s, z]);
+            pairs.push(tup.into_py(py));
+        }
+
+        let qi = PyModule::import_bound(py, "qiskit.quantum_info")?;
+        let spo_cls = qi.getattr("SparsePauliOp")?;
+        let list_bound = PyList::new_bound(py, pairs);
+        let spo = spo_cls.call_method1("from_list", (list_bound.as_any(),))?;
```
kevinsung, line 517: We might as well just have this function return the list directly. `ffsim.qiskit.jordan_wigner` can call `SparsePauliOp` itself. Then, we won't need to import `SparsePauliOp` in Rust.

Another request I have is, can we rewrite this function so that we use `SparsePauliOp.from_sparse_list` instead of `from_list`? The input is more compact that way. See https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.quantum_info.SparsePauliOp#from_sparse_list.

davidomanovic: Yes sure, I can fix it tomorrow! I could make a new folder in src called qiskit with a function jordan_wigner.rs, or should I just make a standalobe function inside fermion_operator.rs? I'll import from ffsim._lib into the python file and only do python stuff here. I agree with the Sparse Pauli list of course!

kevinsung: Great, thanks! Yeah, I think it makes sense to create a new file called `jordan_wigner.rs`, but for now you can just put it at the top level in `src/` rather than creating a `qiskit` subdirectory.

davidomanovic: Have a look at the new pull request https://github.com/qiskit-community/ffsim/pull/484! The `SparsePauliOp.from_sparse_list` gave even more of a speed boost when evaluating energy 😄

davidomanovic:
This PR is superceded by #484.

Issue #496, opened 2025-11-01
Add Rayon multithreading in `jordan_wigner.rs`


davidomanovic:
Btw @davidomanovic if this can be multithreaded in Rust (for example using Rayon) that would be a welcome contribution too. I just didn't want to have Python-level multiprocessing for now.

_Originally posted by @kevinsung in https://github.com/qiskit-community/ffsim/pull/484#issuecomment-3474469842_

PR #497, opened 2025-11-01
Feature: JW map Rayon Multithreading

davidomanovic:
# Description
Multithreading in Rayon for faster JW map by parallelizing processes in Rust.

# Results

On N2 in cc-pvdz:

1 thread: 14.098538247868419 s
4 threads: 5.772801402956247 s
16 threads: 3.734854843467474 s
48 threads: 3.3065379448235035 s
64 threads: 3.1618218813091516 s

Closes issue https://github.com/qiskit-community/ffsim/issues/496

### DIFF
--- a/src/jordan_wigner.rs
+++ b/src/jordan_wigner.rs
@@ -3,8 +3,6 @@ use numpy::Complex64;
 use pyo3::prelude::*;
 use rayon::prelude::*;
 use std::collections::HashMap;
-use std::env;
-use std::sync::Once;
 
 type SparseLabel = String;
 type SparseIndices = Vec<usize>;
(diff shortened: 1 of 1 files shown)


kevinsung:
Just to be sure, this is more performant than your original Python multiprocessing implementation, right?

davidomanovic:
> Just to be sure, this is more performant than your original Python multiprocessing implementation, right?

Yes, even the non-parallelized Rust implementation was faster than the original Python multiprocessing implementation.

src/jordan_wigner.rs
```diff
@@ -18,104 +21,163 @@ pub fn jordan_wigner_qiskit(
...
-                let phase_x = multiply_by_zs_and_main(&mut s_x, &z_positions, q, b'X');
-                *next.entry(s_x).or_insert(Complex64::new(0.0, 0.0)) += c * coeff_x * phase_x;
-
-                // Y branch
-                let mut s_y = ps;
-                let phase_y = multiply_by_zs_and_main(&mut s_y, &z_positions, q, b'Y');
-                *next.entry(s_y).or_insert(Complex64::new(0.0, 0.0)) += c * coeff_y * phase_y;
+    init_thread_pool(); // Build the global Rayon pool once
+
+    let (sparse_list, n_qubits) = Python::with_gil(|py| {
```
kevinsung: Excuse my ignorance, but could you explain why we need the GIL here, but I didn't need it for the other places in ffsim where I used Rayon? For example, https://github.com/qiskit-community/ffsim/blob/1cfee7a1958d6dcff56a9a9238d842c811a08f89/src/gates/diag_coulomb.rs#L49.

src/jordan_wigner.rs
```diff
@@ -18,104 +21,163 @@ pub fn jordan_wigner_qiskit(
...
+
+            Ok::<_, ()>((sparse_list, n_qubits))
+        })
+    })
+    .map_err(|_| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>("JW mapping failed"))?;
 
     Ok((sparse_list, n_qubits))
 }
+
+// Initialize the global Rayon thread-pool once, using env if provided.
```
kevinsung: Excuse my ignorance again, but what happens if we don't do this? Is it just for performance?

src/jordan_wigner.rs
```diff
@@ -18,104 +21,163 @@ pub fn jordan_wigner_qiskit(
...
+        // 2) OMP_NUM_THREADS (OpenMP-style)
+        // 3) One of the common BLAS-style env vars
+        // 4) Default (Rayon decides)
+        let from_env = |k: &str| env::var(k).ok().and_then(|v| v.parse::<usize>().ok());
+
+        let n_threads = from_env("RAYON_NUM_THREADS")
+            .or_else(|| from_env("OMP_NUM_THREADS"))
+            .or_else(|| from_env("OPENBLAS_NUM_THREADS"))
+            .or_else(|| from_env("MKL_NUM_THREADS"))
+            .or_else(|| from_env("NUMEXPR_NUM_THREADS"));
```
kevinsung: I'd rather not handle any environment variables manually. `RAYON_NUM_THREADS` should work regardless, right? If so, that's good enough.

src/jordan_wigner.rs
davidomanovic: You are completely right, this is my mistake. We dont need GIL since we don't interact with any python objects

src/jordan_wigner.rs
davidomanovic: Rayon actually handles it automatically, which I did not know. This will be removed.

src/jordan_wigner.rs
davidomanovic: Rayon actually handles it automatically, which I did not know. This will be removed.

src/jordan_wigner.rs
davidomanovic, line 1: Accumulating all fermion terms in parallel

PR #501, opened 2025-11-10
Bug fix: Clamp `c` value in `ffim/qiskit/gates/orbital_rotation.py/`

davidomanovic:
Clamp the value of 'c' to the range [-1.0, 1.0] to prevent raising `math.acos(c)` math domain error from floating point errors causing `c` to be outside the range when decomposing `OrbitalRotationJW` gate.

- Closes issue https://github.com/qiskit-community/ffsim/issues/502

### DIFF
--- a/python/ffsim/qiskit/gates/orbital_rotation.py
+++ b/python/ffsim/qiskit/gates/orbital_rotation.py
@@ -196,6 +196,11 @@ def _orbital_rotation_jw(
 ) -> Iterator[CircuitInstruction]:
     givens_rotations, phase_shifts = linalg.givens_decomposition(orbital_rotation)
     for c, s, i, j in givens_rotations:
+        c = float(c)
+        if c > 1.0:
+            c = 1.0
+        elif c < -1.0:
+            c = -1.0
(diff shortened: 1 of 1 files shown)


tests/python/qiskit/gates/orbital_rotation_test.py
```diff
@@ -211,3 +212,27 @@ def test_inverse_spinless(norb: int, nocc: int):
...
+        return (
+            [(1.0 + 1e-12, 1.0, 0, 1)],
+            [1.0, 1.0],
+        )
+
+    monkeypatch.setattr(
+        ffsim.qiskit.gates.orbital_rotation.linalg,
+        "givens_decomposition",
+        givens_decomposition_fp,
+    )
```
kevinsung: This is overriding the `givens_decomposition` function though. Do you have an example where that function returns an invalid matrix?

davidomanovic: Yeah it was a bad test, I reverted the test script. No idea how to structure it to be honest.

Issue #512, opened 2025-11-27
Bug: Panic in `apply_givens_rotation_in_place` for some active spaces


davidomanovic:
I am seeing a Rust panic when applying UCJOpSpinBalanced to a Hartree Fock state for C2H4 with a 12 orbital active space. Happens for all basis sets.

The same setup works if I reduce the active space to 10 orbitals.

The panic is:
```
thread '<unnamed>' panicked at src/gates/orbital_rotation.rs:79:45:
slice index starts at 258 but ends at 252
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::slice::index::slice_index_order_fail::do_panic::runtime
   3: core::slice::index::slice_index_order_fail
   4: std::thread::scoped::scope
   5: ffsim::gates::orbital_rotation::apply_givens_rotation_in_place
   6: ffsim::gates::orbital_rotation::__pyfunction_apply_givens_rotation_in_place
   7: pyo3::impl_::trampoline::trampoline
   8: ffsim::gates::orbital_rotation::<impl ffsim::gates::orbital_rotation::apply_givens_rotation_in_place::MakeDef>::_PYO3_DEF::trampoline
   9: cfunction_call
  10: _PyObject_MakeTpCall
  ...
pyo3_runtime.PanicException: slice index starts at 258 but ends at 252
```

The Hilbert space dimension matches the expected combinatorial dimension, so it does not look like I am passing inconsistent (norb, nelec).

For my failing case I get `len(Phi0) = 853776, dim_expected = 853776`.

# Minimal reproducer
This should run instantly and raise the bug
```
import numpy as np
import pyscf
import pyscf.gto, pyscf.scf, pyscf.cc
import ffsim
from math import comb

mol = pyscf.gto.Mole()
mol.build(
    atom=[
        ["C", (0, 0, 0.5)],
        ["C", (0, 0, -0.5)],
        ["H", (0, 0.9289, 0.5 + 0.5626)],
        ["H", (0, -0.9289, 0.5 + 0.5626)],
        ["H", (0, 0.9289, -0.5 - 0.5626)],
        ["H", (0, -0.9289, -0.5 - 0.5626)],
    ],
    basis="cc-pvtz",
    symmetry="d2h",
    verbose=0,
)

scf = pyscf.scf.RHF(mol).run()

ncore = 2
ncas = 12
active_space = range(ncore, ncore + ncas)
norb = len(active_space)

[...]

PR #513, opened 2025-11-27
Fix slice index bug in orbital_rotation.rs

davidomanovic:
### Problem
There is a subtle bug in the parallell part of `apply_givens_rotation_in_place`. When `n_pairs` is not a multiple of `n_threads`, the last few chunks can get `start > end` and cause a panic when slicing the `slice1_slice[start..end]` range.

### Fix
if `start >= n_pairs` for a given k, break the loop before slicing.

Added also a test case that fails without the proposed fix.

- Closes issue #512.


kevinsung (approved):
Thanks for catching this!

PR #517, opened 2025-12-12
Add projection-based fast Slater sampler

### LINKED ISSUE #297: Improve Slater determinant sampler performance
The algorithm introduced in #257 spends most of its time computing many determinants. There are more efficient algorithms described in the following references:

- https://arxiv.org/abs/2210.05871
- https://arxiv.org/abs/2109.07358
- https://arxiv.org/abs/1207.6083

q-inho:
# Summary
- Add a projection-based `sample_slater` Slater determinant sampler implementing the fast fermion sampling (FFS) algorithm of Sun, Zou, and Li, https://arxiv.org/abs/2109.07358 (2023) ("Fermion sampling made more efficient").
- Retain the existing determinant-based autoregressive sampler as a robust fallback and as a reference implementation (need discussion to decide retaining it or not)
- Extend the Slater sampler test suite to cover projector vs non-projector cases
- Resolve #297 

## Algorithm 1: Projection-based fast sampling from a Slater determinant specified by occupied orbitals and an orbital rotation
**Goal.**
$\quad$ Generate $S$ independent samples of an $N$-electron occupation bitstring on $L$ orbitals, using the projection-based method with Gaussian-elimination updates.
$\quad$ If the provided 1-RDM is not a Hermitian projector, abort the fast path

**Inputs.**
$\quad$ $\Gamma$: $L \times L$ one-body reduced density matrix
$\quad$ $N$: target electron number
$\quad$ $S$: number of shots (samples)
$\quad$ $RNG$: a pseudorandom number generator

**Outputs**
$\quad$ Sample $\{b^{(s)}\}_{s=1}^S$, each $b^{(s)}\in\{0,1\}^L$ with 
```math
\sum_{i=0}^{L-1} b_i^{(s)}= k.
```

### Pseudocode for implementation
image
image

## Test added
- Same test cases for both `sample_slater` as `sample_slater_determinant` test cases

### 1. Experimental setup
Benchmarks `sample_slater` against `sample_slater_determinant` method. Benchmarks run in the M4 macbook air with 16Gb Memory. 

### 2. Sweeps
Four cases are benchmarked and plotted:
- **Top-left (spinless, half filling):** A spin-polarized Slater determinant on ($L$) orbitals with ($N=L/2$) fermions is sampled for ($L=8,12,\dots,64$) that the wall-clock time per independent configuration is compared between the projection-based sampler and the determinant-based autoregressive sampler, using 500 configurations per data point and taking the median over 5 repeats.
- **Top-right (spinful, spin-balanced half filling):** A spinful Slater determinant with separate projector 1-RDMs for ($\alpha$) and ($\beta$) sectors is sampled with ( $N_\alpha=N_\beta=L/2$ ) for the same set of ($L$). Timing per independent configuration is compared between the two samplers with the same 500-sample, median-of-5 protocol.
- **Bottom-left (spinless, fixed particle number):** A spin-polarized Slater determinant with fixed ($N=8$) is sampled while increasing the orbital count ($L=16,24,\dots,80$), isolating the dependence on system size at fixed fermion number. Each point uses 500 configurations and the median over 7 repeats.
- **Bottom-right (spinful, fixed particle number):** A spinful Slater determinant with fixed ( $(N_\alpha,N_\beta)=(4,4) $) is sampled over the same ($L=16,24,\dots,80$) sweep, again isolating size dependence at fixed fermion number. Each point uses 500 configurations and the median over 7 repeats.

### 3. Results
image

### DIFF
--- a/python/ffsim/states/sample_slater.py
+++ b/python/ffsim/states/sample_slater.py
@@ -470,13 +470,15 @@ def _sample_from_projection_normals(
     """
     norb, nelec = normals.shape
     perm = rng.permutation(nelec)
-    h = normals[:, perm].copy()
+    # Working set of normals after applying the random permutation.
+    active_normals = normals[:, perm].copy()
     selected: list[int] = []
     active = nelec
     # Iterate until all electrons are placed
     while active:
-        h0 = h[:, 0]
-        row_norms = np.abs(h0) ** 2
+        # Pivot column used for sampling and elimination.
+        pivot_normal = active_normals[:, 0]
+        row_norms = np.abs(pivot_normal) ** 2
         total = float(np.sum(row_norms))
         if total <= 0:
             raise ValueError("Failed to compute sampling probabilities.")
@@ -486,13 +488,16 @@ def _sample_from_projection_normals(
         if active == 1:
             # No further elimination needed
             break
-        pivot_val = h0[orb]
+        pivot_val = pivot_normal[orb]
         if np.abs(pivot_val) <= tol:
             raise ValueError("Numerical pivot breakdown in fast sampler.")
         # Update remaining normals via Gaussian elimination (O(N^2))
         for j in range(1, active):
-            h[:, j] = h[:, j] - (h[orb, j] / pivot_val) * h0
+            active_normals[:, j] = (
+                active_normals[:, j]
+                - (active_normals[orb, j] / pivot_val) * pivot_normal
+            )
         # Drop the first normal vector
-        h = h[:, 1:active]
+        active_normals = active_normals[:, 1:active]
         active -= 1
     return sum(1 << orb for orb in selected)


q-inho:
@kevinsung Thank you for the feedback. However since I am still not confident for next steps, I would like to ask for your confirmation for todo lists:
- Intorudce `ffsim.sample_slater(norb, occupied_orbitals, orbital_rotation=None, ...)` that match `ffsim.slater_determinant` inputs with only proposed fast method.
- Compute the the 1-RDM internally from `(occupied_orbitals, orbital_rotation)` reuse the existing sampling implementation.
- Keep the current 1-RDM-based entrpy point `ffsim.sample_slater_determinant` for backward compatibility but mark it with deprecation warning and point users to `sample_slater` that original old code (`determinant` method only).

Does this align with your expectation? Let me know whether I understand your comment correctly.

python/ffsim/states/sample_slater.py
```diff
@@ -39,6 +227,10 @@ def sample_slater_determinant(
 ) -> Sequence[int] | Sequence[str] | np.ndarray:
     """Collect samples of electronic configurations from a Slater determinant.
 
+    Note:
```
kevinsung: Instead of adding a note, use the `deprecated` decorator from `typing_extensions`. There aren't currently any deprecated functions in ffsim, but you can see https://github.com/qiskit-community/ffsim/pull/338 for examples of the usage before they were removed.

python/ffsim/states/sample_slater.py
```diff
@@ -211,3 +403,96 @@ def _generate_probs(rdm: np.ndarray, sample: set[int], norb: int) -> np.ndarray:
         indices = list(sample | {orb})
         probs[i] = np.linalg.det(rdm[indices][:, indices]).real
     return probs / np.sum(probs)
+
+
+class _FastProjectionSampler:
```
kevinsung: This is really great, but in ffsim we actually prefer a functional programming style when possible. Do you think it's sensible to refactor this to use helper functions instead of a helper class while retaining identical performance? I think it would look like this:
- Define a function that computes the normal vectors
- Define a sampling function that takes the normal vectors as input, together with other needed information like `norb` and `nelec`

python/ffsim/states/sample_slater.py
```diff
@@ -211,3 +403,96 @@ def _generate_probs(rdm: np.ndarray, sample: set[int], norb: int) -> np.ndarray:
...
+        """
+        if orbitals.ndim != 2:
+            raise ValueError("Projector orbitals must be a 2D array.")
+        self._norb, self._nelec = orbitals.shape
+        if self._nelec == 0:
+            raise ValueError("Fast sampler requires at least one occupied orbital.")
+        # Precompute normal vectors h_t from QR
+        q, r = np.linalg.qr(orbitals, mode="reduced")
+        diag = np.abs(np.diag(r))
+        tol = 1e-12
```
kevinsung: Generally, tolerances like this should be passed as an argument to the function, with a default value if appropriate.

python/ffsim/states/sample_slater.py
```diff
@@ -211,3 +403,96 @@ def _generate_probs(rdm: np.ndarray, sample: set[int], norb: int) -> np.ndarray:
...
+            rng: NumPy Generator to use for randomness.
+
+        Returns:
+            Integer-encoded bitstring for the sampled configuration.
+        """
+        perm = rng.permutation(self._nelec)
+        h = self._normals[:, perm].copy()
+        selected: list[int] = []
+        active = self._nelec
+        tol = 1e-12
```
kevinsung: Same comment about tolerances being passed to functions.

python/ffsim/states/sample_slater.py
```diff
@@ -211,3 +403,96 @@ def _generate_probs(rdm: np.ndarray, sample: set[int], norb: int) -> np.ndarray:
...
+        # Iterate until all electrons are placed
+        while active:
+            h0 = h[:, 0]
```
kevinsung: If you can think of descriptive names for `h` and `h0`, that would be preferable to single-letter names.

q-inho: I renamed `h` and `h0` to `active_normal` and `pivot_normal`. I expect this could describe that `active_normal` is current working matrix and `pivot_normal` is the leading column used for Gaussian elimination.

PR #518, opened 2025-12-13
Qiskit gates: Support MCPhaseGate

q-inho:
Part of #369 
## Summary
`MCPhaeGate` acts on `m` control qubits and one target qubit, and is diagonal in the computational basis.
It applies a phase only to the basis state where the controls equal `c` and the target equals `1`
If $x=c$ and $t=1$:
```math
U\lvert x, t\rangle = e^{i \theta}\lvert x,t\rangle
```
Otherwise
```math
U\lvert x,t\rangle=\lvert x,t\rangle
```
Equivalently,
```math
U = I + (e^{i\theta}-1)\lvert c,1\rangle\langle c,1\rvert
```
### Discussion point (Need discussion)
- Qiskit’s `MCPhaseGate` is often realized via a decomposition into elementary gates, so finite precision simulation can leave tiny residual amplitudes (roundoff) that do not cancel exactly, which shows up as a small mismatch against ffsim when comparing statevectors.
- One option is to loosen `np.testing.assert_allclose` tolerances. 
- Instead, this PR tentatively rewrites MCPhaseGate instructions into an equivalent DiagonalGate when constructing the Qiskit reference (`_rewrite_for_stable_reference`), which uses an explicit diagonal representation and avoids the decomposition-induced numerical drift. This may be changed after discussion.

### DIFF
--- a/python/ffsim/contract/two_body.py
+++ b/python/ffsim/contract/two_body.py
@@ -61,7 +61,7 @@ def two_body_linop(
     def matvec(vec: np.ndarray):
         result = contract_2e(
             two_body_tensor,
-            vec.astype(complex, copy=False),
+            vec,
             norb,
             nelec,
             link_index=link_index,
@@ -74,7 +74,7 @@ def two_body_linop(
         result = contract_2e(
             # TODO come up with a way to test this transpose
             two_body_tensor.transpose(1, 0, 3, 2).conj(),
-            vec.astype(complex, copy=False),
+            vec,
             norb,
             nelec,
             link_index=link_index,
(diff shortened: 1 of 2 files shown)


kevinsung:
I would just loosen the `np.testing.assert_allclose` tolerance. How large does it need to be? Also, do you think it makes sense to open an issue at Qiskit about this?

kevinsung:
Ahh okay, `1e-12` is actually a pretty small tolerance. In that case opening an issue at Qiskit is probably unnecessary, but I'll leave it to you.

kevinsung (approved):

PR #622, opened 2026-03-28
linalg: move givens_decomposition implementation to Rust

to24toro:
## Summary

Move `ffsim.linalg.givens_decomposition` internals from Python to Rust (PyO3), while keeping the Python interface unchanged.

## What changed

- Added Rust linalg module and Givens decomposition implementation:
  - `src/linalg/mod.rs`
  - `src/linalg/givens.rs`
- Exported new Rust function from `_lib`:
  - `src/lib.rs`
- Switched Python `givens_decomposition` to call Rust backend and preserve return type (`list[GivensRotation], np.ndarray`):
  - `python/ffsim/linalg/givens.py`
- Added typing stub for new `_lib` function:
  - `python/ffsim/_lib.pyi`

## Compatibility

- Python API stays the same for `ffsim.linalg.givens_decomposition`.
- Existing callers continue to receive `GivensRotation` objects and phase-shift diagonal array.

## Comments
- https://github.com/qiskit-community/ffsim/issues/197
Since we now have Rust implementations of zrot and zrotg, it should be possible to integrate them into other related functions as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "maturin"
 [project]
 name = "ffsim"
 requires-python = ">=3.10"
-version = "0.0.73.dev"
+version = "0.0.74.dev"
 description = "Faster simulation of fermionic quantum circuits."
 readme = "README.md"
 license = { file = "LICENSE" }
--- a/python/ffsim/linalg/double_factorized_decomposition.py
+++ b/python/ffsim/linalg/double_factorized_decomposition.py
@@ -1131,3 +1131,51 @@ def _double_factorized_t2_alpha_beta_explicit(
 
 def _quadrature(mat: np.ndarray, sign: int):
     return 0.5 * (1 - sign * 1j) * (mat + sign * 1j * mat.T.conj())
+
+
+def reconstruct_t2(
+    diag_coulomb_mats: np.ndarray, orbital_rotations: np.ndarray, nocc: int
+) -> np.ndarray:
+    """Reconstruct t2 amplitudes from double factorization tensors."""
+    return (
+        1j
+        * contract(
+            "kpq,kap,kip,kbq,kjq->ijab",
+            diag_coulomb_mats,
+            orbital_rotations,
+            orbital_rotations.conj(),
+            orbital_rotations,
+            orbital_rotations.conj(),
+        )[:nocc, :nocc, nocc:, nocc:]
+    )
+
+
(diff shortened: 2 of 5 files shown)


src/linalg/givens.rs
```diff
@@ -0,0 +1,163 @@
...
+use pyo3::exceptions::PyValueError;
+use pyo3::prelude::*;
+
+#[derive(Clone, Copy)]
+struct GivensRotation {
+    c: f64,
+    s: Complex64,
+    i: usize,
+    j: usize,
+}
```
kevinsung: Let's get rid of this struct, it seems unnecessary now. We can also get rid of the corresponding Python namedtuple subclass, which existed only for the sake of readability.

src/linalg/givens.rs
```diff
@@ -0,0 +1,163 @@
...
+
+type GivensRotationTuple = (f64, Complex64, usize, usize);
+type GivensDecompositionResult = (Vec<GivensRotationTuple>, Py<PyArray1<Complex64>>);
+
+fn is_close_zero(x: Complex64, tol: f64) -> bool {
+    x.norm() <= tol
+}
```
kevinsung: Let's get rid of this helper function, it's not doing much. Just write the code in place.

src/linalg/givens.rs
```diff
@@ -0,0 +1,163 @@
...
+
+fn zrot(x: &mut Array1<Complex64>, y: &mut Array1<Complex64>, c: f64, s: Complex64) {
+    let n = x.len().min(y.len());
+    for idx in 0..n {
+        let x_old = x[idx];
+        let y_old = y[idx];
+        x[idx] = c * x_old + s * y_old;
+        y[idx] = c * y_old - s.conj() * x_old;
+    }
+}
```
kevinsung: `zrot` can actually be implemented in terms of `zdrot` and `zscal`, see https://github.com/qiskit-community/ffsim/blob/df10e79ea8a61ff10730b341de305daf2e49b3fe/src/gates/orbital_rotation.rs#L115-L117. Do you think that might improve performance?

src/linalg/givens.rs
to24toro: Removed `is_close_zero` helper.

src/linalg/givens.rs
to24toro: Removed the Rust `GivensRotation` struct

PR #644, opened 2026-04-15
Modified `givens_decomposition` logic for compact circuit compilation

spiccinelli:
This PR corrects the logic of the `zrotg_safe` function, handling the case where both matrix elements `a` and `b` are close to `0` and therefore return an identity gate. This results in more efficient compilation for circuits close to the identity (as Trotter steps for a small time interval) by eliminating unnecessary SWAP operations. To test this case, a new function called `test_givens_decomposition_near_identity` has been added.

### DIFF
--- a/src/linalg/givens.rs
+++ b/src/linalg/givens.rs
@@ -86,15 +86,13 @@ pub fn givens_decomposition(
             for j in 0..=i {
                 let target_index = i - j;
                 let row = n - j - 1;
-                if current_matrix[[row, target_index]] != Complex64::new(0.0, 0.0) {
+                if current_matrix[[row, target_index]].norm() > tol {
                     let (c, s) = zrotg_safe(
                         current_matrix[[row, target_index + 1]],
                         current_matrix[[row, target_index]],
                         tol,
                     );
-                    if (c - 1.0).abs() > tol {
-                        right_rotations.push((c, s, target_index + 1, target_index));
-                    }
+                    right_rotations.push((c, s, target_index + 1, target_index));
                     rotate_columns_in_place(
                         &mut current_matrix,
                         target_index + 1,
@@ -108,15 +106,13 @@ pub fn givens_decomposition(
             for j in 0..=i {
                 let target_index = n - i + j - 1;
                 let col = j;
-                if current_matrix[[target_index, col]] != Complex64::new(0.0, 0.0) {
+                if current_matrix[[target_index, col]].norm() > tol {
                     let (c, s) = zrotg_safe(
                         current_matrix[[target_index - 1, col]],
                         current_matrix[[target_index, col]],
                         tol,
                     );
-                    if (c - 1.0).abs() > tol {
-                        left_rotations.push((c, s, target_index - 1, target_index));
-                    }
+                    left_rotations.push((c, s, target_index - 1, target_index));
                     rotate_rows_in_place(&mut current_matrix, target_index - 1, target_index, c, s);
                 }
             }
--- a/src/linalg/givens.rs
+++ b/src/linalg/givens.rs
@@ -17,11 +17,11 @@ type GivensRotationTuple = (f64, Complex64, usize, usize);
 type GivensDecompositionResult = (Vec<GivensRotationTuple>, Py<PyArray1<Complex64>>);
 
 fn zrotg_safe(a: Complex64, b: Complex64, tol: f64) -> (f64, Complex64) {
-    if a.norm() <= tol {
-        return (0.0, Complex64::new(1.0, 0.0));
-    }
     if b.norm() <= tol {
         return (1.0, Complex64::new(0.0, 0.0));
+    }
+    if a.norm() <= tol {
+        return (0.0, Complex64::new(1.0, 0.0));
     }
     let abs_a = a.norm();
     let abs_b = b.norm();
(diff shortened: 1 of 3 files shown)


src/linalg/givens.rs
```diff
@@ -86,13 +86,15 @@ pub fn givens_decomposition(
...
                 let row = n - j - 1;
-                if current_matrix[[row, target_index]] != Complex64::new(0.0, 0.0) {
+                if current_matrix[[row, target_index]].norm() > tol {
                     let (c, s) = zrotg_safe(
                         current_matrix[[row, target_index + 1]],
                         current_matrix[[row, target_index]],
                         tol,
                     );
-                    right_rotations.push((c, s, target_index + 1, target_index));
+                    if (c - 1.0).abs() > tol {
```
kevinsung: The previous check on line 89 should make this check unnecessary, right?

src/linalg/givens.rs
```diff
@@ -106,13 +108,15 @@ pub fn givens_decomposition(
...
                 let col = j;
-                if current_matrix[[target_index, col]] != Complex64::new(0.0, 0.0) {
+                if current_matrix[[target_index, col]].norm() > tol {
                     let (c, s) = zrotg_safe(
                         current_matrix[[target_index - 1, col]],
                         current_matrix[[target_index, col]],
                         tol,
                     );
-                    left_rotations.push((c, s, target_index - 1, target_index));
+                    if (c - 1.0).abs() > tol {
```
kevinsung: Same comment as above

tests/python/linalg/givens_test.py
```diff
@@ -58,6 +59,24 @@ def test_givens_decomposition_reconstruct(dim: int):
         np.testing.assert_allclose(reconstructed, mat)
 
 
+@pytest.mark.parametrize("dim", range(6))
+def test_givens_decomposition_near_identity(dim: int):
+    """Test Givens decomposition of a quasi-identity matrix."""
+    rng = np.random.default_rng()
+    worst_case_lengths = (0, 0, 1, 3, 6, 10)
```
kevinsung: Instead of hard-coding the values here, just compute the worst-case length from `dim`.

tests/python/linalg/givens_test.py
```diff
@@ -58,6 +59,24 @@ def test_givens_decomposition_reconstruct(dim: int):
...
+
+    for _ in range(3):
+        dt = tol = 10 ** rng.uniform(-8, -3)
```
kevinsung: Instead of generating these values randomly, test an explicit list of `tols`, say `[1e-15, 1e-12, 1e-9, 1e-6, 1e-3]`. Then, set the `dt` accordingly.

tests/python/linalg/givens_test.py
```diff
@@ -58,6 +59,24 @@ def test_givens_decomposition_reconstruct(dim: int):
...
+        generator = ffsim.random.random_hermitian(dim, seed=rng)
+        mat = expm(-1j * dt * generator)
+        givens_rotations, _ = givens_decomposition(mat, tol=tol)
+
+        assert len(givens_rotations) <= worst_case_lengths[dim], (
+            f"dim={dim}, dt={dt:.3e}, got {len(givens_rotations)} rotations, "
+            f"expected at most {worst_case_lengths[dim]}"
+        )
```
kevinsung: In addition to checking the length, also check that the decomposition reconstructs the matrix. You may want to factor out a function
```python
def reconstruct_orbital_rotation(givens_rotations, phase_shifts) -> np.ndarray:
```
using the code in the preceding test `test_givens_decomposition_reconstruct`

tests/python/linalg/givens_test.py
spiccinelli: Since we want to test the case of "near identity leads to fewer rotations" I thought we might want to keep the tolerance fixed and test for different `dt`. I realized that fixing `dt` and varying the tolerance would just test the worst-case upper bound, which is always valid — and not what we want to actually verify here.

tests/python/linalg/givens_test.py
spiccinelli: I implemented the change and now the assert statement checks the equality up to `tol` (`np.testing.assert_allclose(reconstructed, mat, atol=tol)`), which makes sense given that the algorithm is supposed to give a valid _exact_ decomposition up to the set tolerance.

tests/python/linalg/givens_test.py
```diff
@@ -48,16 +65,37 @@ def test_givens_decomposition_reconstruct(dim: int):
...
         np.testing.assert_allclose(reconstructed, mat)
 
 
+@pytest.mark.parametrize("dim", range(6))
+def test_givens_decomposition_near_identity(dim: int):
+    """Test Givens decomposition of a quasi-identity matrix."""
+    rng = np.random.default_rng()
+    # Worst case: one elimination per subdiagonal entry of the unitary
+    worst_case_length = (dim * (dim - 1)) // 2
```
kevinsung:
```suggestion
    worst_case_length = dim * (dim - 1) // 2
```
Nit: unnecessary parentheses

tests/python/linalg/givens_test.py
```diff
@@ -48,16 +65,37 @@ def test_givens_decomposition_reconstruct(dim: int):
...
+    tol = 1e-12
+    dts = [1e-1, 1e-3, 1e-6, 1e-9]
+
+    for dt in dts:
+        generator = ffsim.random.random_hermitian(dim, seed=rng)
+        mat = expm(-1j * dt * generator)
+        givens_rotations, phase_shifts = givens_decomposition(mat, tol=tol)
+
+        assert len(givens_rotations) <= worst_case_length, (
```
kevinsung: We should change the `<=` to `<` now, right? However, I'm not sure the test will pass in its current form, do we need `dt` to be smaller than `tol`?

kevinsung (approved):

PR #646, opened 2026-04-16
Swap the order of the `a` and `b` checks in `zrotg_safe`

spiccinelli:
Splits #644: with reference to that PR, this PR just swaps the order of the `a` and `b` checks in `zrotg_safe`; no test is implemented as the `push` logic in `zrotg_safe` is still independent of whether a certain Givens rotation happens to be the identity.

### DIFF
--- a/tests/python/linalg/givens_test.py
+++ b/tests/python/linalg/givens_test.py
@@ -17,7 +17,6 @@ from pathlib import Path
 import numpy as np
 import pytest
 from scipy.linalg.lapack import zrot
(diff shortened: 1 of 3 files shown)


tests/python/linalg/givens_test.py
```diff
@@ -17,6 +17,7 @@
 import numpy as np
 import pytest
 from scipy.linalg.lapack import zrot
+from scipy.linalg import expm
```
kevinsung:
```suggestion
```

PR #671, opened 2026-07-13
Gate Optimization in DiagCoulombEvolutionJW

jamie-k-kim:
Changed the gate decomposition in the number operator representation (`_diag_coulomb_evo_num_rep_jw`). Right now it yields PhaseGates and CPhaseGates for every diagonal Coulomb term. This new approach accumulates all single-qubit Z-rotations into a coefficient array (`z_coeffs`), applies them all at the start using a single PhaseGate per qubit, and then uses RZZGates for the two-qubit interaction terms.

By the Jordan-Wigner mapping for number operators, $n_p = \frac{I-Z_p}{2}$ and the interaction term $n_p n_q$ expands to $\frac{I - Z_p - Z_q + Z_p Z_q}{4}$. So for single-qubit Z terms, each term $Z_{pq} n_p n_q$ contributes a factor of $\frac{t Z_{pq}}{4}$ to the $Z_p$ and $Z_q$ phases. These are accumulated via `coef = time * this_mat[i, j] / 4.0`. And for two-qubit ZZ terms, each term contributes $\exp\left(-i \frac{t Z_{pq}}{4} Z_p Z_q\right)$, which is equivalent to `RZZGate(0.5 * time * Z_pq)`.

For a 50-orbital system with 100 trials:

| | Avg. CX count | Avg. RZ count | Avg. depth
| -- | -- | -- | -- |
| **Current** | 4883.04 | 4883.04 | 1541.07 |
| **New** | 4842.04 | 4842.04 | 1414.03 |

Transpilation speedup: 24.9% (MacBook Pro M1)

### DIFF
--- a/python/ffsim/qiskit/gates/diag_coulomb.py
+++ b/python/ffsim/qiskit/gates/diag_coulomb.py
@@ -190,39 +190,69 @@ def _diag_coulomb_evo_num_rep_jw(
     else:
         mat_aa, mat_ab, mat_bb = mat
 
-    # gates that involve a single spin sector
+    z_coeffs = np.zeros(2 * norb)
+
+    # Accumulate coefficients for gates that involve a single spin sector
     for sigma, this_mat in enumerate([mat_aa, mat_bb]):
         if this_mat is not None:
             for i in range(norb):
                 if this_mat[i, i]:
-                    yield CircuitInstruction(
-                        PhaseGate(-0.5 * this_mat[i, i] * time),
-                        (qubits[i + sigma * norb],),
-                    )
+                    z_coeffs[i + sigma * norb] += 0.5 * time * this_mat[i, i] * 0.5
+            for i, j in _iter_pairs(norb):
+                if this_mat[i, j]:
+                    coef = time * this_mat[i, j] / 4.0
+                    z_coeffs[i + sigma * norb] += coef
+                    z_coeffs[j + sigma * norb] += coef
+
+    # Accumulate coefficients for gates that involve both spin sectors
+    if mat_ab is not None:
+        for i in range(norb):
+            if mat_ab[i, i]:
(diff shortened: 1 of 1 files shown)


kevinsung:
Thanks for the pull request!

I prefer the current implementation because it maps more directly on to the fermion evolution. Your approach seems like it should give the same number of gates. Do you have an explanation for the ~40 gate difference that you observe? Can you provide a simple example code to reproduce it?

jamie-k-kim:
Sorry, I pasted the numbers wrong. The point of this was to reduce the RZ count by 35-40%, as CPhaseGate decomposes into 2 CX + 3 RZ but RZZGate decomposes into 2 CX + 1 RZ. The only difference is the decomposition, the unitaries themselves are identical. No worries if you prefer the current implementation, I just thought this was worth flagging.

[Here's the script](https://github.com/user-attachments/files/30143548/compare_decompositions.py) so you can verify.

```
Correctness (norb=5):
Circuits are equivalent

Transpilation (norb=50, heavy-hex, 20 runs):
          Avg CX     Avg RZ    Avg depth   Time (s)
old       4948.0     3026.0       1546.0      3.761
new       4906.0     1885.0       1402.0      2.854
```

kevinsung:
Your findings seem to be an artifact of your chosen basis gates. For example, I changed the basis gates in your script to
```
['cz', 'id', 'rz', 'sx', 'x']
```
which is more relevant to IBM's newer generations of QPUs (Heron, Nighthawk) and got
```
Transpilation (norb=50, heavy-hex, 20 runs):
          Avg CZ     Avg RZ    Avg depth   Time (s)
old       5091.0     3785.0       3408.0     43.747
new       5091.0     4868.0       3598.0     44.388
```
If you still prefer RZZGate to CPhaseGate though, you can try setting `z_representation=True` in `DiagCoulombEvolutionJW`.

PR #684, opened 2026-08-03
Add single-layer UCJ energy

hkbelagali:


hkbelagali:
@kevinsung I was looking into why the test cases are getting stuck in CI, I believe it can be narrowed down to a deadlock in jax 0.10.2. This error does not occur on my laptop when I run the test cases, but if I restrict to 4 cores like the GitHub Actions runners, then I am able to reproduce the deadlock in the UCJ algorithm implementation. This code also produces the same deadlock on jax `0.10.2`, but works fine with `0.9.2`.

```py
import time
import numpy as np
import jax
import jax.numpy as jnp

jax.config.update("jax_enable_x64", True)

n_calls = 16
batch = 1024
n  = 5 

rng = np.random.default_rng(0)
A = jnp.asarray(rng.normal(size=(batch, n, n)) + 1j * rng.normal(size=(batch, n, n)))

def f(t):
    return sum(jnp.real(jnp.sum(jnp.linalg.det(A * jnp.exp(1j * (t + k)))))
               for k in range(n_calls))
c = jax.jit(f).lower(0.3).compile()
t = time.time(); o = jax.block_until_ready(c(0.3))

print(f"OK exec {time.time()-t:.2f}s val={float(o):.3f}")
```

```
JAX_PLATFORMS=cpu taskset -c 0-3 python test.py
```

on `jax==0.9.2`, this prints `OK exec 0.01s val=-3054.291`, but it never finishes running on `jax==0.10.2`. I think this is because jax parallelized LAPACK operations in `0.10.0`, and XLA's source code here has a [comment ](https://github.com/openxla/xla/blob/86b98d4bce9a8b452d23642da7e8b14141df9f60/xla/tests/custom_call_test.cc#L653-L656) on the safety of this. I believe the fix for this is also live on the jax main branch right now according to this [PR](https://github.com/jax-ml/jax/pull/39351). The deadlock disappeared when I used a nightly build of jax. Would it be possible to temporarily pin `jax<0.10` until the next release comes out?

kevinsung:
Sure, you can go ahead and edit `pyproject.toml` to restrict to working JAX versions.

hkbelagali:
Sounds good, thanks!

kevinsung (commented):
@hkbelagali Thanks for the contribution! My first request is that you use the newly introduced `rotate_one_body_tensor` and `rotate_two_body_tensor` functions from https://github.com/qiskit-community/ffsim/blob/main/python/ffsim/linalg/util.py. I think these can replace the `_propagate_through_orbital_rotations` and `_propagate_spin_sector_tensor` functions you introduced here. Note, however, that the orbital rotation convention is transposed from your convention (please check this), so you either need to pass `u.T.conj()` everywhere, or rework your logic to align with the ffsim convention.

kevinsung (commented):
To keep things simple for now, let's get rid of the high-level dispatcher functions like `ucj_energy`, `ucj_energy_and_grad`, and `optimize_ucj_energy` and just force the user to use the appropriate function for their operator and Hamiltonian type.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,2125 @@
...
+    Returns:
+        The expectation value of the Hamiltonian with respect to the UCJ state.
+    """
+
+    _validate_ucj_op(ucj_op)
+    _validate_molecular_hamiltonian(hamiltonian, ucj_op.norb)
+
+    pairs_aa, pairs_ab = (
+        interaction_pairs if interaction_pairs is not None else (None, None)
+    )
```
kevinsung: The pairs don't seem to be used later in the function. Is this a bug? If so, we should add a test to catch it.

python/ffsim/variational/ucj_energy.py
kevinsung, line 1: Needs a copyright header.

tests/python/variational/ucj_energy_test.py
kevinsung, line 1: Needs a copyright header.

pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-    "jax",
```
kevinsung, line 26: jax 0.11.1 was just released, let's try reverting this

tests/python/variational/ucj_energy_test.py
```diff
@@ -0,0 +1,336 @@
...
+    )
+    optimized_backprop_energy = ffsim.ucj_energy_spinless(
+        optimized_ucj_op, mol_hamiltonian, nelec
+    )
+    optimized_statevector_energy = _statevector_energy(
+        optimized_ucj_op, mol_hamiltonian, norb, nelec
+    )
+
+    np.testing.assert_allclose(optimized_backprop_energy, result.fun)
+    np.testing.assert_allclose(optimized_backprop_energy, optimized_statevector_energy)
```
kevinsung: Let's simplify these 5 tests into 3 tests that are more consistent. One test for each spin variant. Each test should
- Generate a random UCJ operator (no need to use a molecule)
- Compute the energy and compare it to the state vector reference
- Optimize
- Compute the energy and compare it to the state vector reference
- Assert that the optimized energy is lower than the starting energy
- Be parameterized for a few different values of interaction pairs, such as `[]`, `None`, `[(1, 2), (3, 4)]`

These tests should exercise the default path of not specifying `return_optimize_result`. Then, add separate (cheaper) tests exercising the `return_optimize_result=True` path.

python/ffsim/variational/ucj_energy.py
hkbelagali: Sorry about that! I forgot to remove that code, to my understanding it's not necessary for computing the energy. However, for the optimization runs, we should still need it since we reconstruct UCJ operators.

tests/python/variational/ucj_energy_test.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def test_ucj_energy_chunk_size_nondivisor():
```
kevinsung: Split this into three tests, one for each spin variant.

tests/python/variational/ucj_energy_test.py
```diff
@@ -0,0 +1,336 @@
...
+    chunked = ffsim.ucj_energy_spinless(
+        ucj_op_spinless,
+        mol_hamiltonian_spinless,
+        nelec_spinless,
+        chunk_size=7,
+    )
+    np.testing.assert_allclose(chunked, unchunked)
+
+
+def test_ucj_energy_and_grad():
```
kevinsung: Split this into three consistent tests, one for each spin variant. Extract a helper function to compute the finite difference gradient, and compute the whole gradient instead of just index 0. If this is too expensive, perhaps do it just for a few randomly chosen indices.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1858 @@
...
+            jastrow_mat,
+            jastrow_vec,
+            norb,
+            chunk_size=chunk_size,
+        )
+
+    return jax.jit(jax.value_and_grad(energy, argnums=0))
+
+
+def _spin_balanced_jastrow_phase(same, diff, norb):
```
kevinsung: Please add type annotations to all function signatures, including the helper functions.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1858 @@
...
+    )
+    pairs_aa, pairs_ab = (
+        interaction_pairs if interaction_pairs is not None else (None, None)
+    )
+    validate_interaction_pairs(pairs_aa, ordered=False)
+    validate_interaction_pairs(pairs_ab, ordered=False)
+    if pairs_aa is None:
+        pairs_aa = triu_indices
+    if pairs_ab is None:
+        pairs_ab = triu_indices
```
kevinsung: Since these lines are duplicated between `ucj_energy_and_grad_spin_balanced` and `optimize_ucj_energy_spin_balanced`, extract them to a shared helper function. Do the same for the all three spin variants.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1858 @@
...
+        jac=True,
+        callback=callback,
+        options=options,
+    )
+
+
+def _interaction_pairs_key(
+    interaction_pairs: Sequence[tuple[int, int]],
+) -> tuple[tuple[int, int], ...]:
+    """Functools requires arguments to be hashable."""
```
kevinsung:
```suggestion
    """functools.cache requires arguments to be hashable."""
```
and similarly elsewhere

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1858 @@
...
+            g_pqrs,
+            jastrow_mat,
+            jastrow_vec,
+            norb,
+            chunk_size=chunk_size,
+        )
+    )
+
+
+def ucj_energy_and_grad_spin_balanced(
```
kevinsung: I noticed that this function isn't used in the guide, and it's also not used in `optimize_ucj_energy_spin_balanced`. Let's replace both `ucj_energy_and_grad_spin_balanced` and `optimize_ucj_energy_spin_balanced` with a single function that returns the callable that can be passed to scipy. Something like
```python
def ucj_energy_and_grad_func_spin_balanced(
    hamiltonian, nelec, *, norb, interaction_pairs=None,
    with_final_orbital_rotation=False, occupied_orbitals=None, chunk_size=None,
) -> Callable[[np.ndarray], tuple[float, np.ndarray]]:
```
Then, the guide should use this function to construct the callable, and pass it `scipy.optimize.minimize`. This gives the user more control over the optimization and lets us remove some more code from this file.

Same for the other spin variants.

kevinsung (commented):
I wonder if the variable names with single letters like `q`, `h`, and `g` can be made more descriptive. If you can't think of better names, it's fine though.

tests/python/variational/ucj_energy_test.py
hkbelagali: Since we combined `optimize_...` and `value_and_grad` methods and now just pass a Callable to the user to pass into scipy/other optimizers , do we still need the test cases for the scipy optimize result flag `return_optimize_result`?

kevinsung: no, sorry I wrote this comment before writing the other one!

kevinsung: btw check out https://github.com/qiskit-community/ffsim/blob/main/CONTRIBUTING.md for running the checks locally and installing the pre-commit hooks, it might help with the CI failures

hkbelagali: I'll set up the pre-commit hooks, thank you!

hkbelagali:
> I wonder if the variable names with single letters like q, h, and g can be made more descriptive. If you can't think of better names, it's fine though.

I was trying to match symbols from the paper's equations/lemmas in the code, but I'll change the names to be more descriptive.

tests/python/variational/ucj_energy_test.py
kevinsung: there seems to be a pre-existing unrelated CI failure, I will look into it

kevinsung:
the lint CI failure should be fixed after merging main

tests/python/variational/ucj_energy_test.py
```diff
@@ -0,0 +1,402 @@
...
+    return (plus_energy - minus_energy) / (2 * eps)
+
+
+@pytest.mark.parametrize(
+    "pairs",
+    [
+        None,
+        (None, None),
+        ([(0, 0), (1, 1)], [(0, 0), (1, 1)]),
+        ([(0, 0), (1, 1)], None),
```
kevinsung, line 72: Also need to add an example including the empty list `[]`. Same for the other spins. I think this will trigger an error in `real_symmetrics_from_parameters_jax`.

docs/how-to-guides/simulate-ucj.ipynb
kevinsung, line 1: Run the notebook starting from a freshly initialized kernel and commit it with the outputs.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1442 @@
...
+    *,
+    occupied_orbitals: tuple[Sequence[int], Sequence[int]] | None = None,
+    chunk_size: int | None = None,
+) -> float:
+    """Compute the UCJ energy for a spin-balanced system
+    using the fermionic backpropagation outlined in https://arxiv.org/abs/2607.21337.
+
+    Args:
+        hamiltonian: The Hamiltonian.
+        ucj_op: The UCJ operator. Must have n_reps=1, with an optional final
```
kevinsung, line 55: Order of documented arguments should match function signature (`ucj_op` comes first)

docs/how-to-guides/simulate-ucj.ipynb
```diff
@@ -0,0 +1,294 @@
...
+    "import pyscf\n",
+    "import pyscf.cc\n",
+    "import scipy\n",
+    "\n",
+    "import ffsim\n",
+    "from ffsim.variational.ucj_energy import (\n",
+    "    ucj_energy_and_grad_func_spin_balanced,\n",
+    "    ucj_energy_and_grad_func_spin_unbalanced,\n",
+    "    ucj_energy_spin_balanced,\n",
+    "    ucj_energy_spin_unbalanced,\n",
```
kevinsung, line 38: Don't import these from the private module. Just call them as `ffsim.X`

docs/how-to-guides/simulate-ucj.ipynb
```diff
@@ -0,0 +1,294 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c63b8602",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "info = defaultdict(list)\n",
```
kevinsung, line 142: This is never used, and it's the only reason `defaultdict` is imported.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1442 @@
...
+        jastrow_mat,
+        jastrow_vec,
+        norb,
+        1,
+        chunk_size,
+    )
+    return term_alpha_alpha + term_beta_beta + term_alpha_beta + term_beta_alpha
+
+
+def _compute_energy_spin_balanced(
```
kevinsung, line 1212: Would it make sense to decorate these functions with
```
@jax.jit(static_argnames=["norb", "chunk_size"])
```
? I noticed that the functions that only compute the energy don't use `jit`, and wonder if this would speed them up. I realize this would give nested jits in the energy + grad functions, and I also wonder if there's a way to avoid that (it may not matter though).

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1442 @@
...
+    occupied_beta = tuple(occupied_orbitals[1])
+    if len(occupied_alpha) != nelec[0] or len(occupied_beta) != nelec[1]:
+        raise ValueError(
+            "occupied_orbitals should contain nelec[0] alpha orbitals and "
+            "nelec[1] beta orbitals."
+        )
+    if any(orb < 0 or orb >= norb for orb in occupied_alpha):
+        raise ValueError("Alpha occupied orbital indices are out of range.")
+    if any(orb < 0 or orb >= norb for orb in occupied_beta):
+        raise ValueError("Beta occupied orbital indices are out of range.")
```
kevinsung, line 569: Let's also check that there are no duplicate orbitals in `occupied_alpha` or `occupied_beta`. Same for `_occupied_orbitals_key_spinless`.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1442 @@
...
+    spin: int,
+    chunk_size: int | None,
+) -> jax.Array:
+    """Evaluate all canonical same-spin two-body terms."""
+    n_terms = (norb * (norb - 1) // 2) ** 2
+    if n_terms == 0:
+        return jnp.array(0.0)
+    p_all, q_all, r_all, s_all = (
+        jnp.asarray(array) for array in _canonical_same_spin_two_body_indices(norb)
+    )
```
kevinsung, line 1061: I think we can avoid constructing the `O(N^4)` size tables of `_canonical_same_spin_two_body_indices` by instead doing something like
```
pairs_rows, pairs_cols = jnp.triu_indices(norb, k=1)
```
and then in the `term_chunk` function do
```
        left, right = jnp.divmod(indices, n_pairs)
        p = pair_rows[left]
        r = pair_cols[left]
        q = pair_rows[right]
        s = pair_cols[right]
```
where
```
n_pairs = norb * (norb - 1) // 2
```

Can you verify if this is correct and worth doing?

tests/python/variational/ucj_energy_test.py
hkbelagali: I added the empty list, it first throws an error in the random_ucj_op_spin<> construction since the test cases use these:
```
        if interaction_pairs is None:
            interaction_pairs = (None, None)
>       pairs_aa, pairs_ab = interaction_pairs
        ^^^^^^^^^^^^^^^^^^
E       ValueError: not enough values to unpack (expected 2, got 0)

python/ffsim/random/random.py:613: ValueError
```

After this, the next error occurs in `real_symmetrics_from_parameters_jax`. Should I add the guard at the top of the UCJ energy calculation functions, or to the random UCJ constructors and `real_symmetrics_from_parameters_jax`?

python/ffsim/variational/ucj_energy.py
hkbelagali: I didn't use `jit` with the functions that only compute energy because I thought these would probably only be used once in a script, and the compilation overhead would dominate. But I'm not familiar with how `jit` works under the hood, I'll do some benchmarks to see how adding `jit` to the inner functions and energy functions will affect the runtime.

tests/python/variational/ucj_energy_test.py
kevinsung: I think the correct test case is `([], [])`, not `[]`. After that, what's the error? For now, add the guard in the UCJ energy functions, not `real_symmetrics_from_parameters_jax`. I will think about editing `real_symmetrics_from_parameters_jax` myself.

hkbelagali: Ah, sorry I misunderstood. It would be `([], [])`, `([], [], [])`, and `[]` for the balanced, unbalanced, and spinless operators respectively right? After making these changes, the error is in `real_symmetrics_from_parameters_jax` for the spin balanced and unbalanced versions, so I can add guards in the UCJ energy functions. For the spinless version, the error is in the random UCJ constructor:
```
        # Zero out diagonal coulomb matrix entries if requested
        if interaction_pairs is not None:
            mask = np.zeros((norb, norb), dtype=bool)
>           rows, cols = zip(*interaction_pairs)
            ^^^^^^^^^^
E           ValueError: not enough values to unpack (expected 2, got 0)
```
After this, the spinless version also errors on `real_symmetrics_from_parameters_jax`.

python/ffsim/variational/ucj_energy.py
hkbelagali: I ran the benchmarks, and adding the decorators does seem to help with runtime, thanks! Does it get faster despite the overhead because the operations get fused together after compilation?

tests/python/variational/ucj_energy_test.py
kevinsung: In https://github.com/qiskit-community/ffsim/pull/708 I made the helper functions and the random UCJ constructor accept empty lists, so you can simplify now.

hkbelagali: Sounds good, thanks!

python/ffsim/variational/ucj_energy.py
kevinsung: Yes, I think it's due to the fusing.

python/ffsim/variational/ucj_energy.py
hkbelagali: Yes I think this works, thanks! I implemented it and tested on some systems and memory usage drops by a good amount.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1430 @@
...
+        occupied_orbitals_key,
+        chunk_size,
+    )
+
+    one_body_tensor = jnp.asarray(hamiltonian.one_body_tensor)
+    two_body_tensor = jnp.asarray(hamiltonian.two_body_tensor)
+    constant = jnp.asarray(hamiltonian.constant)
+
+    def scipy_func(x: np.ndarray) -> tuple[float, np.ndarray]:
+        value, grad = value_and_grad(
```
kevinsung, line 194: We should validate the parameter vector length before passing it through to JAX, otherwise when there is no final orbital rotation, excess parameters are silently ignored. You can compute the expected number of parameters like this:
```python
    expected_n_params = UCJOpSpinBalanced.n_params(
        norb,
        n_reps=1,
        interaction_pairs=interaction_pairs,
        with_final_orbital_rotation=with_final_orbital_rotation,
    )
```
Then define a helper function to validate that `x` is 1d and has the correct length. Call the helper function first thing in `scipy_func`.

Same for the other two cases.

python/ffsim/variational/ucj_energy.py
```diff
@@ -0,0 +1,1430 @@
...
+    n_occ = occ_coeffs.shape[1]
+    norb = occ_coeffs.shape[0]
+    phase_factors = jnp.exp(1j * diagonal_phases)
+    phased_coeffs = phase_factors[:, :, None] * occ_coeffs[None, :, :]
+    overlap = jnp.einsum("pi,bpj->bij", occ_coeffs_conj, phased_coeffs)
+    det = jnp.linalg.det(overlap)
+    overlap_rhs = jnp.broadcast_to(
+        occ_coeffs_conj.T, (diagonal_phases.shape[0], n_occ, norb)
+    )
+    transition_density = phased_coeffs @ jnp.linalg.solve(overlap, overlap_rhs)
```
kevinsung, line 950: The calculation here is unstable when `overlap` is close to being singular. I used GPT to find this issue and implement a fix in https://github.com/kevinsung/ffsim/commit/50dbd7e314a2896533fa99d3ce4e96d74847ed8b. The fixed code avoids using `solve` and seems to be faster. Would you mind reviewing this change to see if it makes sense, and if it is a good change, incorporate it into your PR?

Here is GPT's full writeup, including a repro for the instability causing large energy error:

### Problem

For a diagonal phase matrix `D` and occupied-orbital coefficient matrix `Q`, the
implementation forms the occupied-space overlap

```text
M = Q† D Q
```

and then computes

```python
det = jnp.linalg.det(overlap)
transition_density = phased_coeffs @ jnp.linalg.solve(overlap, overlap_rhs)
```

The energy expressions subsequently multiply the transition density by
`det(overlap)`. Algebraically, these products can remain finite when `M` is singular:

```text
det(M) M⁻¹ = adj(M).
```

The implementation nevertheless calculates `det(M)` and `M⁻¹` separately. A
singular `M` makes `solve` undefined, while a nearly singular `M` produces very large
intermediate values that are expected to cancel against a very small determinant.
That cancellation is numerically unstable. The same problem is more pronounced for
the two-body contraction, which contains products of two inverse-derived transition
densities.

`D` being unitary does not guarantee that `Q† D Q` is invertible: the phase-rotated
occupied subspace can be partly or completely orthogonal to the original occupied
subspace. These are valid UCJ parameter values, and an optimizer can also pass close
to them.

### Observed impact

A spinless example with four orbitals and two electrons reproduces the issue. Use a
normalized 4-by-4 Hadamard matrix for the orbital rotation, and choose a real
symmetric Jastrow matrix such that one two-body transition has phase factors
approximately `(1, 1, 1, -1)`. The corresponding occupied-space overlap has rank one.

The following is a self-contained reproduction against the reviewed, solve-based
implementation:

```python
import numpy as np

import ffsim

norb = 4
nelec = 2

# Q is formed from the first two columns of this Hadamard rotation. For one of
# the two-body transitions below, Q† D Q has rank one.
orbital_rotation = 0.5 * np.array(
    [
        [1, 1, 1, 1],
        [1, -1, 1, -1],
        [1, 1, -1, -1],
        [1, -1, -1, 1],
    ],
    dtype=complex,
)
diag_coulomb_mat = (np.pi / 6) * np.array(
    [
        [0, -1, 1, -2],
        [-1, 0, 1, -2],
        [1, 1, 0, 2],
        [-2, -2, 2, 0],
    ],
    dtype=float,
)
ucj_op = ffsim.UCJOpSpinless(
    diag_coulomb_mats=diag_coulomb_mat[None],
    orbital_rotations=orbital_rotation[None],
)

# Move a tiny distance from the exactly rank-deficient point. The physical
# energy is smooth, but det(M) and solve(M, ...) become badly scaled.
params = ucj_op.to_parameters()
params[0] -= 1e-10
ucj_op = ffsim.UCJOpSpinless.from_parameters(
    params,
    norb=norb,
    n_reps=1,
)

hamiltonian = ffsim.random.random_molecular_hamiltonian_spinless(
    norb,
    seed=0,
)
reference = ffsim.slater_determinant(norb, range(nelec))
state = ffsim.apply_unitary(
    reference,
    ucj_op,
    norb=norb,
    nelec=nelec,
)
linop = ffsim.linear_operator(hamiltonian, norb=norb, nelec=nelec)
statevector_energy = np.vdot(state, linop @ state).real
backpropagated_energy = ffsim.ucj_energy_spinless(
    ucj_op,
    hamiltonian,
    nelec,
)

print(f"state vector:    {statevector_energy:.16f}")
print(f"backpropagation: {backpropagated_energy:.16f}")
print(f"absolute error:  {abs(backpropagated_energy - statevector_energy):.3e}")
```

Before the bordered-determinant fix, this prints:

```text
state vector:    -1.1225081045966403
backpropagation: -1.1173267624850860
absolute error:  5.181e-03
```

With the fix described below, the same reproduction gives
`-1.1225081045966392` and `-1.1225081045966385`, respectively, an absolute error
below `1e-15`.

At the singular point, the backpropagated energy can happen to agree with the
state-vector result because of floating-point rounding. Tiny valid perturbations show
the instability clearly. In one tested random Hamiltonian:

| Parameter perturbation | State-vector energy | Backpropagated energy | Error |
| --- | ---: | ---: | ---: |
| `-1e-6` | `-1.1225086523792236` | `-1.1223683782355460` | `1.40e-4` |
| `-1e-10` | `-1.1225081045966403` | `-1.1173267624850860` | `5.18e-3` |

The autodifferentiated gradient is also unreliable near this point. For example, one
Jastrow parameter whose state-vector finite-difference derivative is zero returned a
gradient of approximately `-0.0703`. In exact-singular cases, backend-dependent
`NaN` or infinite values are also possible.

This affects all three public implementations because spin-balanced,
spin-unbalanced, and spinless evaluation all call `_transition_batch`.

### Recommended fix

Do not represent the determinant-weighted contractions as a determinant multiplied by
an independently computed inverse. Compute the finite polynomial quantities directly.
The downstream code needs three distinct objects:

1. The overlap `det(M)`.
2. The determinant-weighted one-body transition density,
   `det(M) * rho`.
3. The determinant-weighted antisymmetrized two-body transition density,
   `det(M) * (rho[q, p] * rho[s, r] - rho[s, p] * rho[q, r])`.

The one-body object can be formed with the adjugate:

```text
det(M) rho = (D Q) adj(M) Q†.
```

The two-body object can similarly be formed from second-order cofactors. A convenient
way to obtain both quantities without explicitly constructing cofactor tensors is to
use bordered determinants. Define

```text
X = D Q
Y = Q†
M = Y X.
```

For selected orbital rows `a` and columns `b`, the Schur-complement identity gives

```text
-det([[M, Y[:, b]], [X[a, :], 0]])
    = det(M) rho[a, b].
```

For two selected rows `(a, c)` and columns `(b, d)`, it gives

```text
det([[M, Y[:, (b, d)]], [X[(a, c), :], 0]])
    = det(M) * (rho[a, b] * rho[c, d] - rho[a, d] * rho[c, b]).
```

Unlike the Schur-complement derivation, the determinant on the left is still defined
when `M` is singular. It is exactly the cofactor polynomial needed by the matrix
element. The bordered matrix is only one or two rows and columns larger than `M`, so
its determinant has the same cubic asymptotic cost as the current determinant and
solve.

### Concrete implementation sketch

Replace `_transition_batch` with helpers that return only quantities which remain
defined at singular overlaps:

```python
def _transition_factors(
    diagonal_phases: jax.Array,
    occ_coeffs: jax.Array,
) -> tuple[jax.Array, jax.Array, jax.Array]:
    """Return M = Q† D Q, X = D Q, and Y = Q† for a phase batch."""
    q_dag = jnp.conj(occ_coeffs).T
    phase_factors = jnp.exp(1j * diagonal_phases)
    phased_coeffs = phase_factors[:, :, None] * occ_coeffs[None, :, :]
    overlap = jnp.einsum("ip,bpj->bij", q_dag, phased_coeffs)
    return overlap, phased_coeffs, q_dag


def _overlap_batch(
    diagonal_phases: jax.Array,
    occ_coeffs: jax.Array,
) -> jax.Array:
    """Return <Q|D|Q> for each diagonal phase in a batch."""
    overlap, _, _ = _transition_factors(diagonal_phases, occ_coeffs)
    return jnp.linalg.det(overlap)


def _det_weighted_transition_batch(
    diagonal_phases: jax.Array,
    occ_coeffs: jax.Array,
    row_orbitals: jax.Array,
    col_orbitals: jax.Array,
) -> jax.Array:
    """Return determinant-weighted one- or two-body transitions.

    ``row_orbitals`` and ``col_orbitals`` both have shape ``(batch, k)``.
    ``k=1`` returns ``det(M) * rho[row, col]``. ``k=2`` returns the
    determinant-weighted antisymmetrized product of the two selected rows and
    columns.
    """
    overlap, phased_coeffs, q_dag = _transition_factors(
        diagonal_phases, occ_coeffs
    )
    batch_size = diagonal_phases.shape[0]
    body_order = row_orbitals.shape[1]

    # X[(a, c), :] for each member of the batch: (batch, k, n_occ).
    selected_rows = phased_coeffs[
        jnp.arange(batch_size)[:, None], row_orbitals
    ]

    # Y[:, (b, d)] for each member of the batch: (batch, n_occ, k).
    # Advanced indexing first produces (n_occ, batch, k).
    selected_cols = jnp.transpose(q_dag[:, col_orbitals], (1, 0, 2))

    zeros = jnp.zeros(
        (batch_size, body_order, body_order), dtype=overlap.dtype
    )
    upper = jnp.concatenate([overlap, selected_cols], axis=2)
    lower = jnp.concatenate([selected_rows, zeros], axis=2)
    bordered = jnp.concatenate([upper, lower], axis=1)

    # det([[M, Y], [X, 0]]) = (-1)^k det(M) det(X M^-1 Y).
    return (-1) ** body_order * jnp.linalg.det(bordered)
```

The helper above was checked against the existing `det * rho` and
`det * antisymmetrized(rho * rho)` expressions on random, well-conditioned complex
inputs. The maximum discrepancy was approximately `1.1e-16`, so the index ordering
and the `(-1) ** body_order` sign agree with the current implementation away from the
singular case.

This implementation also naturally handles zero- and one-electron sectors. For
example, a two-body transition with fewer than two occupied orbitals produces a
rank-deficient bordered determinant and therefore zero, without special-casing an
inverse of a 0-by-0 or 1-by-1 overlap.

The current one-body call site can then change from

```python
det_sector, rho_sector = _transition_batch(phi[:, sector_slice], q_sector)
det_other, _ = _transition_batch(phi[:, other_slice], q_other)

return jnp.sum(
    h_sector[p, q]
    * const
    * det_sector
    * det_other
    * rho_sector[rows, q, p]
)
```

to

```python
weighted_transition = _det_weighted_transition_batch(
    phi[:, sector_slice],
    q_sector,
    row_orbitals=q[:, None],
    col_orbitals=p[:, None],
)
det_other = _overlap_batch(phi[:, other_slice], q_other)

return jnp.sum(h_sector[p, q] * const * weighted_transition * det_other)
```

The same-spin two-body call site should request a second-order bordered determinant:

```python
weighted_wick = _det_weighted_transition_batch(
    phi[:, sector_slice],
    q_sector,
    row_orbitals=jnp.stack([q, s], axis=1),
    col_orbitals=jnp.stack([p, r], axis=1),
)
det_other = _overlap_batch(phi[:, other_slice], q_other)

return 0.5 * coeff * const * weighted_wick * det_other
```

For opposite-spin terms, each spin sector contributes a determinant-weighted
one-body transition, so no separate determinant factors are needed:

```python
weighted_left = _det_weighted_transition_batch(
    phi[:, _spin_slice(left_spin, norb)],
    q_left,
    row_orbitals=q[:, None],
    col_orbitals=p[:, None],
)
weighted_right = _det_weighted_transition_batch(
    phi[:, _spin_slice(right_spin, norb)],
    q_right,
    row_orbitals=s[:, None],
    col_orbitals=r[:, None],
)

return 0.5 * g_flat[indices] * const * weighted_left * weighted_right
```

The spinless one-body and two-body paths use the same substitutions. For the
spinless two-body path, use rows `stack([q, s])` and columns `stack([p, r])`, matching
the existing Wick contraction.

The bordered-determinant path can be used unconditionally: it removes `solve`, keeps
the same asymptotic scaling, and avoids a condition-number branch whose threshold
would introduce another numerical decision into the differentiated objective. If
benchmarking shows that a well-conditioned `solve` path is materially faster, retain
it only as an optimization and select the bordered path for ill-conditioned batches.
A diagonal regularizer or pseudoinverse is not a correct fallback because it changes
the finite cofactor limit at exactly singular overlaps.

The code should verify JAX gradients at exactly rank-deficient bordered matrices. JAX
currently produces cofactor-based derivatives for `jnp.linalg.det`, but a local
custom-JVP determinant based on minors would be the fallback if a supported backend
returns non-finite determinant derivatives. Returning an unweighted `rho` should be
avoided because it inherently requires division by a possibly zero overlap.

### Tests to add

- A structured spinless UCJ case whose occupied-space overlap is rank deficient,
  compared against state-vector energy evaluation.
- Small perturbations on both sides of the singular point, again compared against the
  state-vector result.
- Gradient comparisons near the singular point, using finite differences of the
  state-vector energy rather than finite differences of the same backpropagation
  implementation.
- Chunked and unchunked variants, to ensure the fallback is independent of batching.
- At least one spinful case, since the spinful paths combine overlap factors from two
  sectors differently.

### Implemented fix and measured performance

The working tree now implements the bordered-determinant helpers and updates every
spinless and spinful call site described above. It also adds
`test_ucj_energy_singular_overlap_spinless`, which contains the structured
rank-deficient example, checks chunked and unchunked energies against state-vector
evaluation, and compares five autodifferentiated gradient entries with finite
differences of the state-vector energy.

The regression improves the previously observed `5.18e-3` energy error to less than
`1e-15`. The five sampled gradient discrepancies are at most approximately `2.4e-9`;
the solve-based implementation had errors as large as `7e-2` in the same case.

Performance was measured on the JAX CPU backend with JAX 0.11.1, NumPy 2.5.3, and an
x86-64 Intel Xeon Cascadelake processor. Both implementations were loaded in the same
Python process: the original implementation came from `HEAD`, while the working-tree
module contained the bordered-determinant implementation. After compiling both, the
benchmark alternated old and new calls and reports median synchronous wall time. The
four-orbital cases used `chunk_size=16` and 20 samples per callable.

| Four-orbital case | Operation | Solve-based | Bordered determinant | Change |
| --- | --- | ---: | ---: | ---: |
| Spinless, 2 electrons | Energy | 2.568 ms | 2.404 ms | -6.4% |
| Spinless, 2 electrons | Energy + gradient | 1.522 ms | 1.305 ms | -14.3% |
| Balanced, (2, 2) electrons | Energy | 9.292 ms | 7.408 ms | -20.3% |
| Balanced, (2, 2) electrons | Energy + gradient | 11.462 ms | 6.654 ms | -41.9% |
| Unbalanced, (2, 2) electrons | Energy | 10.891 ms | 9.563 ms | -12.2% |
| Unbalanced, (2, 2) electrons | Energy + gradient | 9.585 ms | 5.947 ms | -38.0% |

A second warmed benchmark used six orbitals, `chunk_size=64`, and 10 samples:

| Six-orbital case | Operation | Solve-based | Bordered determinant | Change |
| --- | --- | ---: | ---: | ---: |
| Spinless, 3 electrons | Energy | 3.532 ms | 2.876 ms | -18.6% |
| Spinless, 3 electrons | Energy + gradient | 4.530 ms | 3.017 ms | -33.4% |
| Balanced, (3, 3) electrons | Energy | 30.633 ms | 19.032 ms | -37.9% |
| Balanced, (3, 3) electrons | Energy + gradient | 67.691 ms | 32.020 ms | -52.7% |

Separate-process first-call measurements, which include JAX tracing and compilation,
varied between a 2% improvement and a 15% regression depending on the ansatz. That
range is noisy and compilation dominates it; it does not indicate a material
steady-state regression. The paired warmed results consistently favor the new path.
This is plausible because each Hamiltonian term now computes only the selected scalar
transition via one bordered determinant rather than solving for the entire `norb` by
`norb` transition-density matrix.

kevinsung (commented):
Thanks for your patience @hkbelagali ! We are getting close. One more small change, as well as a numerical instability found by GPT which I would like you to review.

PR #707, opened 2026-09-11
qiskit/gates: introduced Fermionic FFT with Cooley-Tukey alg.

### LINKED ISSUE #665: Add fermionic fast Fourier transform

novitoll:
Resolves: https://github.com/qiskit-community/ffsim/issues/665

Reference: https://www.emergentmind.com/topics/fermionic-fast-fourier-transform


python/ffsim/qiskit/gates/fermionic_fft.py
```diff
@@ -0,0 +1,150 @@
...
+    def inverse(self):
+        return FermionicFFTSpinlessJW(self.norb).conjugate()
+
+
+def _fermionic_fft_jw(qubits: Sequence[Qubit]) -> Iterator[CircuitInstruction]:
+    """Yield instructions executing a Cooley-Tukey fermionic FFT on the given qubits.
+
+    For powers of 2, uses an efficient Cooley-Tukey decomposition requiring only
+    log2(N) distinct angle classes for twiddle factors / Givens rotations.
+    For arbitrary N, falls back to standard Givens decomposition of the Fourier matrix.
```
kevinsung, line 103: Thank you very much for the contribution!

Let's go further and implement a Cooley-Tukey algorithm that works for arbitrary composite integers by performing a prime factorization. For each prime size, emit an OrbitalRotation gate.

For the prime factorization, just write a simple trial division implementation for now.

novitoll: @kevinsung , thanks for the review. Could you add more clarifications? I'm a bit confused here. 
Do you mean that the currently for arbitrary N whch is not a power of 2, we fall back to full N x N
DFT matrix in Givens decomposition which is not efficient with O(N^2) gate params,
and by factoring out N into prime components, e.g. N = 6 = 2 x 3, we can use these small sub-matrices instead? 

Is it then a [prime-factor FFT algorithm](https://en.wikipedia.org/wiki/Prime-factor_FFT_algorithm) or is it still Cooley-Tukey algorithm with the optimization of using prime factors?

kevinsung: Currently, you apply the Cooley-Tukey algorithm only when N is a power of 2. But in fact, the Cooley-Tukey algorithm can be generalized so that it works for arbitrary integers, as described here: https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm#Idea. The base case is when you need to perform a DFT for a prime size. In that case, we just use an OrbitalRotation gate.

novitoll: @kevinsung , PTAL. I've made the generic Cooley-Tukey impl. from the Wikipedia, also removed `inverse()` method, I guess, the default https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.Instruction#inverse should be enough for now. 

```bash
$ pytest tests/python/qiskit/gates/fermionic_fft_test.py
======================================================== test session starts =========================================================
platform linux -- Python 3.13.11, pytest-9.1.1, pluggy-1.6.0
...
configfile: pyproject.toml
collected 39 items                                                                                                                   

tests/python/qiskit/gates/fermionic_fft_test.py .......................................                                        [100%]

========================================================= 39 passed in 2.17s =========================================================

```

Sorry for the big delay. Should have more focused time now.  If there's a Discord, Slack, where we could chat with IM, please let me know

